PR #3, opened 2017-08-15
[Monorepo] Adopt @glimmer/component and @glimmer/application

tomdale:
This PR begins the process of turning this repo into a monorepo that contains "application-level" Glimmer packages (as opposed to "VM-level" packages in [glimmer-vm](https://github.com/glimmerjs/glimmer-vm)).

To start, we import `@glimmer/component` and `@glimmer/application`. These are imported using the script in `script/import-packages.sh` which preserves each repository's git history.

### Lerna for managing dependencies

Each package can have its own dependencies; these are installed by running `lerna bootstrap`.

### Single Root `tsconfig.json`

This prevents VS Code from getting confused when one package imports another across `tsconfig.json` boundaries, which may have different settings. The single `tsconfig.json` keeps VS Code happy and also keeps code quality consistent across subpackages.

### Use `paths` for package resolution

We use `tsconfig.json`'s `paths` setting to allow TypeScript to find local packages, even when imported via package name. This is preferable to symlinking into `node_modules` because relying on Node module resolution causes TypeScript to look at `package.json`, which points at `.d.ts` files instead of the actual TypeScript source. The implication of this is that when you command-click a symbol from another package in VS Code, you get taken to the actual source code instead of a generated `.d.ts` file, and cross-package changes are reflected in the editor immediately without waiting for a build step.

### Unified build

Build and test all packages in the repository via a single Broccoli build script. The build infrastructure is very similar to the recent work for glimmer-vm with some tweaks:

1. Work with existing repository layout, so we can import without major changes
2. Run tests in ES6+ instead of transpiling ES5 as the VM does since we have less stringent compatibility constraints and this allows us to do things like use async functions in tests.


build/broccoli/build-tests.js
```diff
@@ -0,0 +1,154 @@
...
+  let transpiled = babel(merge([simpleHTMLTokenizer, simpleDOM]), {
+    sourceMaps: 'inline',
+    moduleIds: true,
+    plugins: [
+      'transform-es2015-modules-amd'
+    ]
+  });
+
+  let glimmerVM = funnel('packages', {
+    include: ['@glimmer/*/node_modules/@glimmer/*/dist/amd/es5/*.js']
```
rwjblue: I don't fully grok what is happening here. What creates `packages/@glimmer/<whatever>/node_modules/@glimmer/compiler`?  What prevents duplicates from being concatted?

build/broccoli/build-tests.js
```diff
@@ -0,0 +1,154 @@
...
+    include: ['index.html']
+  });
+
+  let loaderPath = path.parse(require.resolve('loader.js'));
+  let loader = funnel(loaderPath.dir, {
+    files: [ loaderPath.base ],
+    destDir: '/assets'
+  });
+
+  let qunit = funnel(path.join(require.resolve('qunitjs'), '..'), {
```
rwjblue: Is this going to end up watching all of `node_modules`?

package.json
```diff
@@ -0,0 +1,50 @@
+{
+   "name": "glimmer.js",
+   "version": "0.8.0",
```
rwjblue: Seems odd to bump this here (others are on 0.7.x I believe)

build/symlink-dependencies/index.js
```diff
@@ -0,0 +1,66 @@
...
+ * tree-sync to efficiently copy the output from Broccoli's tmp directory to
+ * dist. Unfortunately, these optimizations are incompatible with symlinks (it
+ * can't handle circular links) so the symlinking must happen after the build
+ * process has finished.
+ */
+module.exports = {
+  name: 'symlink-dependencies',
+
+  preBuild() {
+    rimraf('dist/@glimmer/*/node_modules');
```
rwjblue: This seems pretty weird: clearing `dist/` when we may not even be building into `dist/`

build/symlink-dependencies/index.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
```
rwjblue: This addon seems unused?  Did you intend to add `ember-addon: { paths: ['./build/symlink-dependencies']}` to the `package.json`?

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: node_js
+sudo: required
+dist: trusty
+cache: yarn
+
+addons:
+  firefox: "54.0"
```
tomdale: 👍

bin/publish.js
```diff
@@ -0,0 +1,206 @@
...
+const DIST_PATH = path.resolve(__dirname, '../dist');
+const PACKAGES_PATH = path.resolve(__dirname, '../packages');
+
+const DRY_RUN = process.argv.indexOf('--dry-run') > -1;
+if (DRY_RUN) {
+  console.log(chalk.yellow("--dry-run"), "- side effects disabled");
+}
+
+// Fail fast if we haven't done a build first.
+assertDistExists();
```
tomdale: Ugh, sorry about that. I updated the script to eagerly perform a build always.

bin/publish.js
```diff
@@ -0,0 +1,206 @@
...
+    if (!stat.isDirectory()) {
+      throwNoPackagesErr()
+    }
+  } catch (e) {
+    throwNoPackagesErr();
+  }
+}
+
+function assertGitIsClean() {
+  let status = execSync('git status').toString();
```
tomdale: TIL, thank you

build/broccoli/build-tests.js
tomdale: Packages are installed here by Lerna during `lerna bootstrap`. Duplicates are indeed concatted, which is allowed by our AMD loader so long as it duplicates are registered before the module is reified. This is obviously not ideal, but requires a mechanism for having per-package dependencies in the tests. I wanted to get something working landed and then iterate on resolving that problem.

build/broccoli/build-tests.js
tomdale: No, because:

```
> require.resolve('qunitjs')
'/Users/tomdale/Code/glimmer.js/node_modules/qunitjs/qunit/qunit.js'
```

So this resolves to `node_modules/qunitjs/qunit`.

package.json
tomdale: I was thinking we should synchronize versions here like we do with glimmer-vm since `@glimmer/application` and `@glimmer/component` are so coupled. They're on 0.7.2 and 0.6.0 respectively now so I think we'd need to bump to 0.8.0 if we want to sync versions.

build/symlink-dependencies/index.js
tomdale: Is line 29 the best way to determine target output directory for the current build?

package.json
rwjblue: Totally agreed that we should synchronize, but what I was actually getting at was that the "bumping" of the version number should be done all at the same time with the publish script (the last time I ran the `bin/publish` in glimmer-vm where one package had a different version number I got an error)

build/symlink-dependencies/index.js
rwjblue: copyToOutputPath

rwjblue: Hmm, not _really_ (it doesn't account for `ember build --output-path=./some-path`) but I don't think we can do better here without changes upstream.  

I'd suggest we add a `process.env.EMBER_CLI_BUILD_OUTPUT_PATH` environment variable (or expose it via some specific API that addons have access to).

build/broccoli/build-tests.js
rwjblue: Sounds good, mind creating an issue and linking to it from a comment?  Just to make sure we don't lose track of the concern...

build/broccoli/build-tests.js
rwjblue: Awesome, thank you for confirming.

chadhietala:
I'm going to block as there is a cycle

```
⠋ BuildingWarning: cycle detected: @glimmer/component <- @glimmer/application <- @glimmer/component
```

PR #15, opened 2017-09-26
Use yarn to install package dependencies

tomdale:


rwjblue:
CI is not happy 😿

tomdale:
@rwjblue Ack, thanks for catching that. I made a booboo and was configuring the extra args as a string when it was supposed to be an array. I think this is good to go now and is much faster than npm, but we should try to switch to Yarn workspaces after we break the circular dependency.

lerna.json
```diff
@@ -3,5 +3,7 @@
   "packages": [
     "packages/@glimmer/*"
   ],
+  "npmClient": "yarn",
+  "npmClientArgs": ["--no-lockfile"],
```
chadhietala: Why?

tomdale: I don’t remember; didn’t we pair on this PR? 🎃 I don’t remember if this is disabled because it caused problems or just to preserve existing Lerna/npm semantics. If we use workspaces then these would get hoisted to the root yarn.lock anyway.

PR #19, opened 2017-09-28
Implement component `bounds` feature

tomdale:
With the recent Glimmer VM changes to adopt IndustryStandardCase, Glimmer.js components now relax the requirement of a template having a single root element. In addition to having multiple elements, this also means that a template may have top-level nodes such as `CommentNodes` or `TextNodes`.

This change renders the previous `element` API incoherent because there is no guarantee of a single "main" element for a template. Instead, we now set the `bounds` property of a component, which contains pointers to the first node and last node.

This is obviously less convenient for cases where you do just have a single conceptual main element. For example, if you have a template with a single root element, you would probably prefer to use `this.element` instead of `this.bounds.firstNode` or `this.bounds.lastNode`.

We could automatically determine the "single element" case and assign element automatically, but we were afraid that it would become a refactoring hazard. If you had a template with a single root element, wrote a bunch of code that used `this.element`, and then someone came along and added a comment to the beginning, all of that code would suddenly stop working.

Instead, you can now indicate which element is the conceptual "main" element by using `...attributes`, which also is the location where invocation-side attributes will get splatted into.
For example:

```hbs
{{!-- MyComponent.hbs --}}
<h1>Title</h1>
<div class="content" ...attributes>
</div>
```

In this example, `this.element` would be the `<div>` element. If you invoked this component with `<MyComponent id="component-id" />`, this `<div>` would also receive the `id` attribute.


rtablada:
@tomdale 👏 The `...attributes` solves an issue I had talked to yahuda about where class name and attribute bindings are lost when using things like data attributes or class names with Ember's current `tagName: ''`.

locks:
Clarifying question: do we need to always set `...attributes`? If so, is it a compile time error if we don't?

tomdale:
@locks No, it is not required. You will get an error if you invoke a component with attributes and it doesn't have splattributes, and I want to add a development-time error that gives helpful guidance if you access `this.element` without `...attributes`. (Working with @chadhietala on making sure we have stripping infrastructure for Glimmer.js.)

tomdale:
@rtablada I don't believe this example works because attribute values can only be strings, and `{{action}}` returns a function. I think in this case we want an element modifier (`<div {{on 'click' myThing}}>`) which IIRC we plan to splat on to `...attributes` as well, though I'm not sure if we have consensus on that.

rtablada:
@tomdale cool. Good work. LMK if there's more discussion on how to handle passing down events.

rtablada:
Just to clarify, I think some confusion will come down to this:

If people expect `...atributes` to pass all native attributes, they will assume `onclick` to be a native attribute.

@tomdale saying that "`...attributes` only passes string attributes" might be clarification enough although I think that there could be a cool story for events too be passed too.

packages/@glimmer/component/src/component.ts
```diff
@@ -126,11 +130,100 @@ import { metaFor } from "./tracked";
...
+   * </div>
+   * ```
+   *
+   * In this example, `this.element` would be the `div` with the class `contents`.
+   *
    * You should not try to access this property until after the component's `didInsertElement()`
    * lifecycle hook is called.
    */
-  element: Simple.Element = null;
+  element: HTMLElement;
```
rwjblue: I don't fully understand this change.  How does this work when rendering in Node-land?

tomdale: Accessing the DOM should only be done after `didInsertElement()`, which isn't invoked in Node-land.

rwjblue: @tomdale - Ya, I don't really agree here.  There are other hooks that are called where `this.element` might be used. A better solution (IMHO) is to pass `element: HTMLElement` in as an argument to `didInsertElement`.  This preserves the proper typings when using `this.element` outside of `didInsertElement`, and allows usage inside `didInsertElement` without casting.

Also, generally speaking, this sort of discussion is super valuable and it is very unfortunate that it could not have happened before the PR was landed...

tomdale:
@rtablada Not sure what you mean by "native" attributes vs. "string" attributes. All attributes are coerced to strings when being set:

```js
let div = document.createElement('div');
div.setAttribute('onclick', function() { });
div.getAttribute('onclick'); // => "function() { }"
typeof div.getAttribute('onclick'); // => "string"
```

cibernox:
The multi root components and `...attributes` are awesome!

I'm seeying on the comments the usafe of `<MyComponent>` instead of `<my-component>`. Is that intentional? Did that change recently?

locks:
Did it? :shipit:

tomdale:
@cibernox Yes, this change was intentional. The use of CapitalComponents (and design of the Glimmer component API in general) dominated the agenda of our recent core team face-to-face meeting. 😄

We will have a write-up soon of the proposed new changes and their rationale. The TL;DR is that they unlock multi-root/fragment components in addition to being less cognitive overhead (you don't have to invent two words when one is the most descriptive).

PR #23, opened 2017-10-12
Fix dependencies and add failing tests for block params and user helpers

chadhietala:
The two tests fail in the same way.


packages/@glimmer/application-test-helpers/yarn.lock
```diff
@@ -0,0 +1,4722 @@
...
+    "@glimmer/util" "^0.29.6"
+
+"@glimmer/resolver@^0.3.0":
+  version "0.3.1"
+  resolved "https://registry.yarnpkg.com/@glimmer/resolver/-/resolver-0.3.1.tgz#41069345b6f41beb0948cc35d8e4aa60adcadfc5"
+  dependencies:
+    "@glimmer/di" "^0.2.0"
+
+"@glimmer/runtime@^0.29.0":
+  version "0.29.6"
```
chadhietala: Needs to be 0.29.7

PR #25, opened 2017-10-17
Introduce @glimmer/compiler-delegates

chadhietala:
This introduces some builtin compiler delegates that can be used in the bundle compiler. A compiler delegate is a class that encapsulates all of the logic for answering questions about the app code on behalf of the bundle compiler. For instance the Module Unification Compiler Delegate encapsulates all the lookup rules for a project using MU. The delegates are meant to be generic in the sense that they are not coupled to a specific build pipeline.


tomdale:
@chadhietala I wonder if it's even worth having the "basic" compiler delegate here at the moment? It's not documented and I'm not sure it even works.

PR #31, opened 2017-10-19
Introduce Application Delegates

chadhietala:
As the Glimmer VM has continued to expand its platform of capabilties it has become clear that supporting a matrix of modes would be very difficult and confusing for users. Instead we should have sane defaults but provide the primitives to changing how a glimmer.js application behaves. Below is a matrix of capabilties that the VM currently supports:

| Mode | Client | Server |
|------|--------|--------|
|ClientSide Rehydration| X | X (Requires rehydration serialization) |
|ClientSide Rendering| X | |
|LazyCompilation| X | X |
|EagerCompilation| X | X |
|StringBuilding (SEO)| | X |

As you can see this matrix is pretty large. To deal with this, this PR introduces the idea of Application Delegates to augment how a glimmer.js application should work. This initial PR first seperates out the destict pieces around rendering.


tomdale:
There's another dimension we probably want to make configurable as well, which is the async behavior of initial render/rehydration. Should it be synchronous, use `requestIdleCallback`, something else?

I'm wondering if it's appropriate to have a single `AppDelegate` or if it's just shuffling deck chairs on the Titanic, as it were. It feels like we may want a delegate for every dimension of configurability?

tomdale:
Superseded by #34.

PR #33, opened 2017-10-22
Restore 0.7.0 this.element behavior

tomdale:
In #19, we added support for `this.bounds` on a component, which allows accessing the first and last nodes of "fragment" templates, i.e. templates without a single root element.

One of the changes in that PR was to how `this.element` works. Instead of `this.element` meaning the root element, it instead meant the attribute with the `...attributes` annotation. Templates without `...attributes` on an element would not have `this.element` set to anything.

In retrospect, this change may have been premature. We do not yet have consensus around the exact semantics of `...attributes`. For example, can you have multiple elements with `...attributes`? In that case, a single property doesn't make sense as the API for accessing the "splattributes" elements.

In order to ease migration from 0.7.x to 0.8.0 for existing Glimmer.js apps, this PR reverts this specific behavior and restores the meaning of `this.element` to be the same as in previous versions: `this.element` is the root element.

Because fragment templates were previously unsupported, there is a small change to behavior. Now, if a fragment component accesses `this.element`, an exception will be thrown with a suggestion to use `this.bounds` instead. This shouldn't affect existing apps because fragment templates would not have compiled. Instead, this is designed to catch errors where users don't realize that a component has multiple root elements.

### DIFF
--- a/packages/@glimmer/component/src/component-manager.ts
+++ b/packages/@glimmer/component/src/component-manager.ts
@@ -92,9 +92,7 @@ export default class ComponentManager implements IComponentManager<ComponentStat
     return new RootReference(bucket.component);
   }
 
-  didCreateElement(bucket: ComponentStateBucket, element: HTMLElement) {
-    bucket.component.element = element;
-  }
+  didCreateElement(bucket: ComponentStateBucket, element: HTMLElement) { }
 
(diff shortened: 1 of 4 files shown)


rwjblue:
I'm in favor of landing this for now with the intent of easing the 0.8 migration, but we need to acknowledge that a future release will almost certainly have larger backwards compatibility issues when we unroll this...

PR #37, opened 2017-10-24
[WIP] Bytecode loader ✈️

tomdale:
This is my flight-with-no-Wi-Fi spike of the bytecode loader. It currently runs a subset of the test suite in both "runtime compiler" and "bytecode" mode. Currently, some helper related tests are failing due a bug in how we resolve helpers.

Mostly, this PR involved a lot of overdue cleanup. The biggest change is that the Application's `boot()` method is now async, to support asynchronous rendering paths that involve either incremental rendering, loading bytecode over the network, or both. Many of the tests were relying on the previously-synchronous behavior, and have been updated accordingly.


packages/@glimmer/application/src/application.ts
```diff
@@ -70,14 +121,26 @@ export default class Application implements Owner {
...
 
   protected _notifiers: Notifier[] = [];
 
   constructor(options: ApplicationOptions) {
     this.rootName = options.rootName;
     this.resolver = options.resolver;
-    this.document = options.document || (typeof window !== 'undefined' && window.document);
+
+    this.document = options.document || DEFAULT_DOCUMENT;
+    this.loader = options.loader || new RuntimeLoader(this.resolver);
```
chadhietala: As we talked about lets remove the defaults, add an assertion, and then we need to update the blueprints.

tomdale:
@chadhietala I think this can be closed now, right?

PR #51, opened 2017-10-27
Optimize template-only components

tomdale:
One of the benefits of template-only components is that they are stateless "pure functions": they operate only on the arguments passed to them. That means that the normal work we have to do around managing state and lifecycle for stateful components can be skipped in template-only components.

Previously, we were instantiating two useless objects per pure component:

1. An empty base Component class (why????)
2. A ComponentStateBucket, which tracks metadata about a particular component instance

In this commit, we instead only instantiate a component and ComponentStateBucket in cases where the user has supplied a Component subclass. In addition to skipping extra allocations, this allows us to bail out early in cases where we do not need to invoke component lifecycle hooks.

As an additional optimization, we can return a single instance of CONSTANT_TAG from getTag(), and return a shared instance of a null ConstReference from getSelf(), reducing the allocations required by stateless components even further and avoiding an opcode execution on updates altogether due to the CONSTANT_TAG.

This commit also addresses a polish/DX issue where component-less templates would produce an opaque error if a developer tried to access properties from the template. Now, accessing a property like `{{foo}}` from a template without a backing component produces a helpful error message suggesting the user may have meant to access a passed argument instead.

In order to produce this error message, we have to allocate a debug-only state bucket for each component instance in development mode that contains information about the invoked template. That means that the above-described optimizations only kick in in production mode, but it didn't seem worth sacrificing helpful error messages. This difference between development mode and production mode leads to some unfortunate boilerplate in order to produce something that will be stripped appropriately, but I couldn't find a better way to do it without sacrificing code size or performance.

### DIFF
--- a/packages/@glimmer/application/test/browser/environment-test.ts
+++ b/packages/@glimmer/application/test/browser/environment-test.ts
@@ -4,6 +4,7 @@ import Component from '@glimmer/component';
 import { buildApp, didRender } from '@glimmer/application-test-helpers';
 import * as SimpleDOM from 'simple-dom';
 import { Simple } from '@glimmer/interfaces';
+import { DEBUG } from '@glimmer/env';
 
 const { module, test } = QUnit;
 const serializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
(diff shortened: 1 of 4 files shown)


packages/@glimmer/application/test/browser/environment-test.ts
```diff
@@ -199,3 +200,29 @@ test('renders a component using simple-dom', async function(assert) {
...
+if (DEBUG) {
+  test('accessing properties in template-only components produces a helpful error in development mode', async function(assert) {
+    assert.expect(1);
+
+    try {
+      await buildApp()
+        .template('Main', '<h1>Hello, {{name}}!</h1>')
+        .boot();
+    } catch (err) {
+      assert.ok(err.message.match("You tried to reference {{name}} from the Main template, which doesn't have an associated component class. Template-only components can only access args passed to them. Did you mean {{@name}}?"));
```
Turbo87: 👍

PR #52, opened 2017-10-27
Bump Glimmer-VM deps to 0.29.10

chadhietala:
This fixed some implicit types around the DOMBuilders. Trying to not introduce any more type errors besides the known issues with simple-dom types. In other news I feel like I want to hook up green keeper if it works on monorepos.


Turbo87:
> I want to hook up green keeper if it works on monorepos.

see https://github.com/greenkeeperio/greenkeeper/issues/139

looks like they have it on the todo list but it doesn't seem supported yet

PR #53, opened 2017-10-29
Update and improve Component documentation

locks:


tomdale:
@locks Looks like Travis is failing due to trailing whitespace.

rwjblue:
@locks  - Ping!

locks:
@tomdale g2g

packages/@glimmer/component/src/tracked.ts
```diff
@@ -12,6 +12,55 @@ import { dict, Dict } from "@glimmer/util";
...
+ * ```typescript
+ * import Component, { tracked } from '@glimmer/component';
+ *
+ * export default class MyComponent extends Component {
+ *    @tracked
+ *    remainingApples = 10
+ * }
+ * ```
+ *
+ * When something changes the component's `remaining` property, the rerender
```
Turbo87: this should be `remainingApples` instead of `remaining`, right?

packages/@glimmer/component/src/tracked.ts
```diff
@@ -12,6 +12,55 @@ import { dict, Dict } from "@glimmer/util";
...
+ * ```typescript
+ * import Component, { tracked } from '@glimmer/component';
+ *
+ * export default class MyComponent extends Component {
+ *   totalApples = 100
+ *
+ *   @tracked
+ *   eatenApples = 0
+ *
+ *   @tracked('totalApples', 'eatenApples')
```
Turbo87: this seems strange. why is `totalApples` mentioned as a dependency, but `totalApples` above is not marked as `@tracked`?

packages/@glimmer/component/src/tracked.ts
```diff
@@ -194,7 +243,7 @@ function hasOwnProperty(obj: any, key: symbol) {
   return hOP.call(obj, key);
 }
 
-let propertyDidChange = function() {};
+let propertyDidChange = function () { };
```
Turbo87: unrelated whitespace changes?

locks: ugh, prettier

packages/@glimmer/component/src/tracked.ts
locks: I think `totalApples` can be removed from the dependencies.

Turbo87: we then need to explicitly explain it though that it does not have to be mentioned because it itself is not `@tracked` and probably even why it is not `@tracked`

packages/@glimmer/component/src/tracked.ts
```diff
@@ -12,6 +12,55 @@ import { dict, Dict } from "@glimmer/util";
...
+ * ```typescript
+ * import Component, { tracked } from '@glimmer/component';
+ *
+ * const totalApples = 100;
+ *
+ * export default class MyComponent extends Component {
+ *    @tracked
+ *    eatenApples = 0
+ *
+ *    @tracked('eatenApples')
```
t-sauer: Might be helpful to provide an example that has multiple dependents.

packages/@glimmer/component/src/tracked.ts
```diff
@@ -12,6 +12,55 @@ import { dict, Dict } from "@glimmer/util";
...
+ * When something changes the component's `remainingApples` property, the rerender
+ * will be scheduled.
+ *
+ * @example Dependents
+ *
+ * In the case that you have a computed property that depends other
+ * properties, you want to track both so that when one of the
+ * dependents change, a rerender is scheduled.
+ *
+ * In the following example we have three properties, `totalApples`,
```
jmbejar: `totalApples` is not a property of the component class. I would change it to:

> In the following example we have two properties,`eatenApples` and `remainingApples`.

Issue #56, opened 2017-10-30
Turn on `noImplicitAny`


chadhietala:
If an application chooses to use the `noImplicitAny` they will have a really bad time with packages coming from this mono repo. While we have done a decent job at typing, we should turn on `noImplicitAny` in the monorepo and fix forward. The largest offender are the test. This issue can be addressed file by file.

## How To Help

1. Locally change `noImplicitAny` in `tsconfig.json` to `true`.
2. Run `yarn run problems`
3. See failures
4. Pick a file to fix the type errors
5. Turn `noImplicitAny` back to false
6. Create PR

We will turn on `noImplicitAny` once we have resolved the issue

c0urg3tt3:
i've done some files, should i PR one by one or can i PR many at once (changes are small) ? 

i picked @glimmer/application cause i saw someone was on @glimmer/app-compiler/src/bundle-compiler, how to track the files done relative to this issue ? maybe a checklist of the file listed in the console output ? let me know ...

givanse:
Updated some: https://github.com/glimmerjs/glimmer.js/pull/148

PR #61, opened 2017-10-31
[wip] partially fix 'Turn on `noImplicitAny` #56'

c0urg3tt3:
i try to use appropriate type when possible, but i mostly use `any` when i'm not sure because it can easily been searched to see what kind of type are really needed later, or while solving this PR.

note: i doesn't touch at `app-compiler/src/bundle-compiler.ts` as @topaxi as submitted a [PR](https://github.com/glimmerjs/glimmer.js/pull/58) relative to this file

### DIFF
--- a/packages/@glimmer/application-test-helpers/src/render-test.ts
+++ b/packages/@glimmer/application-test-helpers/src/render-test.ts
@@ -36,13 +36,14 @@ export class RenderTest {
 }
 
 function setTestingDescriptor(descriptor: PropertyDescriptor): void {
-  let testFunction = descriptor.value as Function;
+  let testFunction = descriptor.value as {[key: string]: any};
   descriptor.enumerable = true;
   testFunction['isTest'] = true;
 }
 
 export interface TestMeta {
   debug?: boolean;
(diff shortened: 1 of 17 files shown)


rwjblue:
@tomdale  - Would you mind starting some initial rounds of review here?

tomdale:
@c0urg3tt3 Thanks for the PR! A couple notes on this:

1. We prefer the `value as Type` casting syntax to the `<Type>value` syntax, as it seems that's the direction the TS team is pushing folks.
2. Generally speaking, I'd prefer to avoid *explicit* `any` as well as implicit `any`. There are some instances where `any` is correct or acceptable (e.g. in tests where properly typing stub objects would be annoying without much value). But there are several instances in this PR where it should be possible to use the proper types. If it's not obvious what they are (which is totally understandable!) maybe we can stick to "real" `any` and other scalar types like `string`, `number` etc for now?

c0urg3tt3:
@tomdale thank you for the advices, i'll try to enhance the PR !

pzuraq:
This has been enabled on master. Thank you for your efforts!

PR #64, opened 2017-11-07
Data Segment Generation Tests

chadhietala:
Prior to this PR we were attempting to do acceptance level assertions on the data segment. This meant we were doing weird things like asserting the length of the content and `indexOf` checks. While these gave some confidence in that we were producing the correct artifact it was extremely fragile. This adds unit tests around the individual pieces of the data segment that gives more confidence for each serialized piece of data.


chadhietala:
This also makes the node tests actually run in CI 🤣 😢

PR #66, opened 2017-11-09
Improve bytecode compilation

tomdale:
This PR is a grab-bag of improvements to the toolchain for compiling applications to Glimmer bytecode.

In particular, this PR:

1. Updates Glimmer VM to 0.30.1.
2. Updates the Module Unification compiler delegate (`@glimmer/compiler-delegates`) to use the new `ModuleLocator`/`TemplateLocator` system for addressing buildtime metadata, and being more rigorous about handling runtime template metadata.
3. Supports a first-class notion of "builtins" in the compiler, allowing Glimmer.js and other extensions to prepopulate the compiler with well-known helpers and templates that can be reified at runtime.
4. Supports encoding type information in the external module table, which allows us to shrink the size of the serialized table considerably.


packages/@glimmer/application/package.json
```diff
@@ -15,21 +15,22 @@
...
-    "@glimmer/runtime": "^0.29.10",
-    "@glimmer/util": "^0.29.10"
+    "@glimmer/runtime": "^0.30.1",
+    "@glimmer/util": "^0.30.1"
   },
   "devDependencies": {
     "@glimmer/application-test-helpers": "^0.9.0-alpha.4",
     "@glimmer/component": "^0.9.0-alpha.4",
-    "@glimmer/wire-format": "^0.29.10",
+    "@glimmer/test-helpers": "^0.30.1",
```
chadhietala: Remove this

packages/@glimmer/application/test/browser/environment-test.ts
```diff
@@ -210,6 +210,7 @@ if (DEBUG) {
         .template('Main', '<h1>Hello, {{name}}!</h1>')
         .boot();
     } catch (err) {
+      console.log(err.message);
```
chadhietala: Remove

packages/@glimmer/application/package.json
tomdale: 👍

packages/@glimmer/application/test/browser/environment-test.ts
tomdale: 👍

PR #67, opened 2017-11-10
Implement Main Invocation For Bytecode Loader

chadhietala:
This introduces the plumbing for the virtual invocation of the `mainTemplate` for a Glimmer Application running with the bytecode format.

# What This PR Does
In lazy mode we have the luxury of having a concrete precompiled `mainTemplate` in javascript that we can invoke by calling `renderLayout` with a preloaded context known as `self`. This would in return bind the seed data and render the UI. In the bytecode mode we do not have a concrete function to invoke, as the template is now just a binary blob. So to begin the execution we need a virtualized way of invoking the `mainTemplate`. To account for this the VM now has a `main` opcode that is responsible for invoking the binary version of the `mainTemplate`. That being said the host needs to do a bit of prep-work before the VM starts executing.

Prior to executing, the environment needs to load the stack with all the pieces of data that are required for a component invocation. This means we need to load an `Invocation`, `ComponentDefinition`, and `ARGS` on the stack in preparation for the invocation of the `mainTemplate`. This forces a lookup convention for data associated with the `mainTemplate`.

A template that is added to the Bundle Compiler with the locator of the following will be used as the `mainTemplate`.

```
{ module: '@glimmer/application', name: 'mainTemplate' }
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,6 @@
     "**/node_modules": true,
     "**/bower_components": true,
     "dist": true
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
 }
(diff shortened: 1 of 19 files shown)


chadhietala:
Closing this for now. But this has the example on how to use the STD Lib Main.

PR #69, opened 2017-11-13
Fix serialization

chadhietala:
This is a more tractable subset of #67 that fixes real issues without trying to align booting an application and booting something that splats components out on the page.

### DIFF
--- a/packages/@glimmer/application/src/loaders/bytecode/resolver.ts
+++ b/packages/@glimmer/application/src/loaders/bytecode/resolver.ts
@@ -30,9 +30,7 @@ export const enum ModuleTypes {
 export interface TemplateLocator {
   module: string;
   name: string;
-  meta: {
-    specifier: string;
-  };
+  specifier: string;
 }
 /**
  * Exchanges VM handles for concrete implementations.
--- a/packages/@glimmer/application/src/loaders/bytecode/resolver.ts
+++ b/packages/@glimmer/application/src/loaders/bytecode/resolver.ts
@@ -30,7 +30,6 @@ export const enum ModuleTypes {
 export interface TemplateLocator {
   module: string;
   name: string;
-  specifier: string;
 }
 /**
  * Exchanges VM handles for concrete implementations.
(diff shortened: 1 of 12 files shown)


packages/@glimmer/application/src/loaders/bytecode/resolver.ts
```diff
@@ -18,6 +18,10 @@ function buildComponentDefinition(ComponentClass: Factory<Opaque>, manager: Comp
   };
 }
 
+export interface TemplateMeta {
```
tomdale: I think this will work in tests but not in real apps. That's because the MU compiler delegate only includes `specifier` information, not the full `locator`. (It's wasteful to include module metadata in the payload that never gets used at runtime.)

Is it possible to change the template meta to just be `specifier: string`? I think that's all the compiler delegate emits.

packages/@glimmer/compiler-delegates/src/module-unification/compiler-delegate.ts
```diff
@@ -39,8 +39,12 @@ export default class MUCompilerDelegate implements AppCompilerDelegate<TemplateM
   }
 
   protected _builtins() {
+    let mainLocator = this.templateLocatorFor({
+      module: '@glimmer/application',
+      name: 'mainTemplate'
+    });
     return {
-      main: helperLocatorFor('main', 'mainTemplate'),
+      main: mainLocator,
```
tomdale: Why is the built-in called `main` but it's called `mainTemplate` everywhere else?

I think we should generally try to avoid special casing the main template and just use the built-ins infrastructure where it exists.

packages/@glimmer/compiler-delegates/src/module-unification/compiler-delegate.ts
```diff
@@ -62,9 +66,15 @@ export default class MUCompilerDelegate implements AppCompilerDelegate<TemplateM
    */
   templateLocatorFor({ module, name }: ModuleLocator): TemplateLocator<TemplateMeta> {
     let relativePath = module.replace(/^\.\//, '');
-    let meta = {
-      specifier: this.project.specifierForPath(relativePath)
-    };
+
+    let meta;
+    if (name === 'mainTemplate') {
```
tomdale: …for example, why are we hardcoding the check for `mainTemplate` here? Shouldn't we just check to see if the locator passed is in `builtins` first and skip normalization in those cases, like we do with helpers?

IOW, the semantics should be: locators in `builtins` are ambient and not normalized, anything else is assumed to come from the project and need normalization.

PR #71, opened 2017-11-14
Share Codegen in Application Builder

chadhietala:
For the bytecode loader we need to be able to refer to the meta data that is created during compilation. This makes a base class of the MU code generator available for in browser tests. To do this we have to stub `@glimmer/compiler-delegates/index.js` to only export the code generator. Otherwise Rollup tries to pull in transitive deps that can not run in the browser.

By doing this we are making sure that the code generation that we do at build time will match up to the runtime code. In other words it attempts to make a single source of truth for codegen. That being said there are two areas that appear to need divergence and that is creating the external module table and specifier generation.

The external module table in the browser tests simply leverages the in memory component classes and the specifier generation is just identity method (just outputs the input).


chadhietala:
Superseded by the end to end smoke test

PR #74, opened 2017-11-15
BUGFIX: Builtin helpers and app helpers serialization

chadhietala:
This fixes the serialization issues related to helpers when running with the bundle compiler. Today there are several types of helpers:

1. Host specific (action, if)
2. App Helpers
3. 3rd Party

When we are serializing the data segment we construct the external module table. The external module table is where the VM trades out an opaque handle for a live JavaScript object. For component classes this works great as they do not need to be wrapped in a reference. Helpers on the other hand do need to be wrapped in a reference prior to evaluation. Further more helpers like `{{action}}` are implemented in a way that produces references where as user space helpers are just functions that get wrapped in references. Since there currently isn't a unification here we need to differentiate in the serialized output if we need to wrap the helper in the a reference. This fixes am issue where this wrapping information was being lost during serialization.

### DIFF
--- a/packages/@glimmer/compiler-delegates/src/module-unification/code-generator.ts
+++ b/packages/@glimmer/compiler-delegates/src/module-unification/code-generator.ts
@@ -182,7 +182,7 @@ export default class MUCodeGenerator {
         relativePath
       );
 
-      return { module: relativePath, name };
+      return {...locator, module: relativePath, name };
     }
   }
 }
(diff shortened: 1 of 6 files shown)


packages/@glimmer/compiler-delegates/src/module-unification/code-generator.ts
```diff
@@ -182,7 +182,7 @@ export default class MUCodeGenerator {
         relativePath
       );
 
-      return { module: relativePath, name };
+      return {...locator, module: relativePath, name };
```
tomdale: Seems like you can omit `name` here (and remove it from the above destructuring assignment altogether).

PR #80, opened 2017-11-23
Remove unused `emberjs-build` dependency

Turbo87:
It seems like this is no longer used anywhere...


rwjblue:
Looks like there are some merge conflicts now (probably yarn.lock?)...

Turbo87:
rebased and fixed

PR #81, opened 2017-11-23
Replace deprecated `qunitjs` dependency with `qunit`

Turbo87:
see https://github.com/qunitjs/qunit/pull/1232

### DIFF
--- a/build/broccoli/build-tests.js
+++ b/build/broccoli/build-tests.js
@@ -109,7 +109,7 @@ function includeTestHarness() {
     include: ['index.html']
   });
 
-  let qunit = funnel(path.join(require.resolve('qunitjs'), '..'), {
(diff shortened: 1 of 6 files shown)


chadhietala:
This created more type errors since the typedefs are not updated.

rwjblue:
Why do type errors not fail the build?

Turbo87:
> This created more type errors since the typedefs are not updated.

where are the typedefs?

rwjblue:
@Turbo87 - Presumably `@types/qunit`?
 
https://github.com/glimmerjs/glimmer.js/blob/master/package.json#L18

rwjblue:
But it doesn't seem like https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/qunit/index.d.ts actually indicates the package that you are requiring/importing from, so I'm unsure what @chadhietala is referring to then...

PR #85, opened 2017-11-23
Remove unused/duplicate dev dependencies

Turbo87:
Out of the three exisiting dev dependencies in `application-test-helpers` one seems unused and the other two are already provided by the root `package.json` file.


Turbo87:
@rwjblue rebased and conflict solved

PR #86, opened 2017-11-23
CI: Run linting step too

Turbo87:
... unfortunately this fails right now because linting is broken 😞


Turbo87:
closing due to inactivity

PR #89, opened 2017-11-28
Refactor Data Segment

chadhietala:
This reduces the overall size of the data segment by doing 2 things:

1) Interning the common prefix of the specifier
2) Collapsing the symbol table and handle map under the same key

To do the interning of the MU specifier we simply collect up all of the specifiers we need to serialize off and then find the common prefix amongst them and make it available to the resolver for resolution.

Prior to these changes the symbol tables and the handle map had the same key (MU specifier). This collapses these 2 concerns into 1 object.

The end result ends up looking something like the following:

```js
const prefix = 'template:/my-project/components/';
const meta = {
  "X": { h: 1, table: { hasEval: 0, symbols: ['@foo'] },
  "Y": { h: 2, table: { hasEval: 0, symbols: ['@bar'] },
...
}

```


tomdale:
We will need to revisit this when we have multiple logical prefixes (via addons etc.) but this is a pragmatic optimization for now that significantly reduces compiled file size. LGTM.

PR #93, opened 2017-12-02
Implicit 'projectPath' for Broccoli Plugin

chadhietala:
This makes the broccoli plugin actually usable within an ember-cli project. Prior to this commit we assuming passing an entire project root into the plugin however in practice just the `src` directory will be used. This moves to an implicit assumption that the `inputPath` is the root of the project. This means the tree that is passed to the plugin must contain the package.json.

### DIFF
--- a/packages/@glimmer/app-compiler/src/bundle-compiler.ts
+++ b/packages/@glimmer/app-compiler/src/bundle-compiler.ts
@@ -41,8 +41,8 @@ export default class GlimmerBundleCompiler extends Plugin {
 
     return Object.assign({
       outputFiles: {
-        heapFile: 'src/templates.gbx',
-        dataSegment: 'src/data-segment.js'
+        heapFile: 'templates.gbx',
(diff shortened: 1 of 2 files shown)


rwjblue:
Not sure how this wouldn’t cause recursive rebuilds on systems without watchman because the project `tmp` would be inside this root tree being passed in, and change events in `./tmp` would trigger a rebuild.

What am I missing here?

rwjblue:
So you are saying we would clone the package.json into an unwatched tree, funnel `src` (as a watched tree), then merge the two together? In other words, we would _not_ be passing in the project root as the root tree, we would just be making it _seem_ like that’s what we are doing?

chadhietala:
@rwjblue Correct. @tomdale Likely can talk more about why the `package.json` needs to be there at all.

tomdale:
The `package.json` is needed to infer the module prefix name.

PR #95, opened 2017-12-07
Use Glimmer.js version of UpdatableReference to avoid action volatility

tomdale:
The UpdatableReference provided by `@glimmer/object-reference` is volatile, and we were inadvertently using it to wrap values produced by the `{{action}}` helper.

This commit uses the version of UpdatableReference from `@glimmer/component`, which is stable until its underlying value is manually updated. This avoids significant performance degradation when values from `{{action}}` helpers are passed as args into other components.

### DIFF
--- a/packages/@glimmer/application/src/helpers/action.ts
+++ b/packages/@glimmer/application/src/helpers/action.ts
@@ -1,4 +1,4 @@
-import { UpdatableReference } from "@glimmer/object-reference";
+import { UpdatableReference } from "@glimmer/component";
 import { VM, Arguments } from "@glimmer/runtime";
 import { Opaque } from '@glimmer/interfaces';
 


packages/@glimmer/application/src/helpers/action.ts
```diff
@@ -1,4 +1,4 @@
-import { UpdatableReference } from "@glimmer/object-reference";
```
rwjblue: Oiy. How many times have we been bitten by this same thing? Any chance of making `@glimmer/object-reference` a private package that doesn’t get published (while we fix glimmer-vm’s test suite to not rely on it)?

tomdale: @rwjblue Yeah, seriously. There is one usage of it left in Glimmer.js, actually, that was not as simple to fix as this one because the circular dependency breaks initialization.

We could fix that in the interim by hoisting the references stuff in `@glimmer/component` into a separate package (`@glimmer/component-reference`?), but I think the real solution is to hoist the reference implementation up into `glimmer-vm` so that it can be shared by both Glimmer.js and Ember.js (we'll need this imminently for interop between the two, anyway).

I know @krisselden had some ideas for a strawman API for creating composable references from multiple environments. I'd like to understand the level of effort for that to decide if we should do something temporary to fix this, or just bite the bullet and build the composable system we'll need soon anyway.

chadhietala:
This is also very bad. https://github.com/glimmerjs/glimmer.js/blob/master/packages/%40glimmer/application/src/helpers/user-helper.ts#L29

PR #96, opened 2017-12-07
Helpers should not be volatile

chadhietala:
When references are combined to produce a single reference a volatile reference poisons the reference effectively making all the references to be re-validated. This removes the volatile references from user land helpers.

### DIFF
--- a/packages/@glimmer/application/src/helpers/user-helper.ts
+++ b/packages/@glimmer/application/src/helpers/user-helper.ts
@@ -5,7 +5,6 @@ import {
 
 import {
   PathReference,
-  Reference,
   TagWrapper,
   RevisionTag
 } from "@glimmer/reference";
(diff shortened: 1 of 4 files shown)


chadhietala:
I'm going to add some tests for this :)

packages/@glimmer/application/src/helpers/user-helper.ts
```diff
@@ -26,10 +26,11 @@ export default function buildUserHelper(helperFunc): GlimmerHelper {
 export class SimplePathReference<T> implements PathReference<T> {
```
tomdale: Can we just use a ConstReference for wrapping user helpers instead of reimplementing PathReference?

chadhietala: This seems to indicate we need a `PathReference` implementation. https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/runtime/lib/environment.ts#L312. I agree this can be better but I think we should land this work first and then do all the type fixes as a fast follow. The fact that these things are volatile is much worst then the duplication.

krisselden: I think there is an PathReference that is const.

krisselden: PrimitiveReference extends ConstReference and implements PathReference, and ValueReference extends PrimitiveReference

PR #97, opened 2017-12-10
Don't use rAF for scheduling re-render

chadhietala:
/cc @krisselden

### DIFF
--- a/packages/@glimmer/application/src/application.ts
+++ b/packages/@glimmer/application/src/application.ts
@@ -178,11 +178,11 @@ export default class Application implements Owner {
 
     this._rendering = true;
     this._scheduled = true;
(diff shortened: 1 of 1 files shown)


krisselden:
@rwjblue we need it so it isn't mutating dom on rAF.

Turbo87:
@krisselden why is that a bad thing? and how is setTimeout improving the situation?

krisselden:
@Turbo87 it fights with stuff that wants to read layout to measure, it is better to batch reads and writes and I believe rAF to be a better place for layout/style reads.  Mutating the dom is essentially the beginning of the next phase of layout paint composite, setTimeout allows for the last phase to complete.

PR #104, opened 2018-01-31
Unify Loader Interface

chadhietala:
This makes getting the `TemplateIterator` have the same interface on the consuming side.

### DIFF
--- a/testem.js
+++ b/testem.js
@@ -15,7 +15,8 @@ let config = {
     "Chrome": [
       "--headless",
       "--disable-gpu",
-      "--remote-debugging-port=9222"
+      "--remote-debugging-port=9222",
+      "--no-sandbox"
     ]
   },
   "launch_in_dev": [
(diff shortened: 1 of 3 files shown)


testem.js
```diff
@@ -15,7 +15,8 @@ let config = {
     "Chrome": [
       "--headless",
       "--disable-gpu",
-      "--remote-debugging-port=9222"
+      "--remote-debugging-port=9222",
+      "--no-sandbox"
```
tomdale: ?

chadhietala: screen shot 2018-01-30 at 5 22 52 pm

rwjblue: This is required for Travis + Chrome usage (the alternative is to use `sudo: required` in your `.travis.yml` which generally increases CI start time)...

tomdale: I don't understand why such a small change (making `getTemplateIterator` async) broke Travis and necessitates a config change? Is this just a change Travis made suddenly?

chadhietala: Looking the build logs the version of chrome is not versioned, it is evergreen. The last successful build was 2 months ago https://travis-ci.org/glimmerjs/glimmer.js/builds/314387954. Also Travis a SAAS and has no semver gurantees. This was recently added to the docs. https://github.com/travis-ci/docs-travis-ci-com/commit/2d65f508acab873efc64665d97342d7da1be34b9

PR #115, opened 2018-02-17
Autotrack, take 2

wycats:
This PR adds "auto-tracking" to tracked getters.

Previously, when defining a tracked computed property, you would write:

```ts
class {
  @tracked first;
  @tracked last;

  @tracked('first', 'last') get name() {
    return `${this.first} ${this.last}`;
  }
}
```

After this change, you can leave off the "dependent keys" and write:

```ts
class {
  @tracked first;
  @tracked last;

  @tracked get name() {
    return `${this.first} ${this.last}`;
  }
}
```

Rather than rely on knowing the full list of dependent properties up front, after this change Glimmer tracks the consumption of all `@tracked` properties inside the getter, and automatically creates a validator based on the consumed keys.

### DIFF
--- a/packages/@glimmer/component/src/component.ts
+++ b/packages/@glimmer/component/src/component.ts
@@ -1,5 +1,6 @@
 import { assert } from "@glimmer/util";
 import { metaFor } from "./tracked";
+import { CURRENT_TAG } from "@glimmer/reference";
 
 export interface Bounds {
   firstNode: Node;
@@ -266,7 +267,7 @@ class Component {
 
   set args(args) {
     this.__args__ = args;
-    metaFor(this).dirtyableTagFor("args").inner.dirty();
+    metaFor(this).updatableTagFor("args").inner.update(CURRENT_TAG);
   }
 
   /** @private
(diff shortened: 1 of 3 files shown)


mmun:
The code looks good to me. I've also written a benchmark here: 

https://github.com/glimmerjs/glimmer.js/compare/autotrack...autotrack-benchmark?expand=1#diff-504d4e1545bcaa7fc620aec828f900b4

The results are great! There's basically no difference between `@tracked` with explicit dependencies and with auto-tracking.

screen shot 2018-02-17 at 2 57 07 am
screen shot 2018-02-17 at 2 56 41 am

locks:
Yay 😍

tomdale:
Also, ideally would you mind updating the body of the PR to explain to viewers at home what's going on here

wycats:
@tomdale done!

sandstrom:
Out of curiosity, would this work too? 😄 

```js
class {
  first: '',
  last: '',

  @tracked get name() {
    return `${this.first} ${this.last}`;
  }
}
```

*In other words, is it enough to tell the computed property that it's tracked, or must all dependent keys be marked as tracked too?*

rwjblue:
Anything not tracked is considered "constant", so in the above if you expect to update on changes from `first` or `last` you would need to track those also...

PR #121, opened 2018-03-19
Fix some noImplicitAny typescript issues

chrisrng:


packages/@glimmer/compiler-delegates/src/module-unification/compiler-delegate.ts
```diff
@@ -4,7 +4,7 @@ import { expect } from '@glimmer/util';
 import { relative } from 'path';
 import { SerializedTemplateBlock } from '@glimmer/wire-format';
 import { CompilableTemplate, CompileOptions, ICompilableTemplate } from '@glimmer/opcode-compiler';
-import Debug from 'debug';
+import Debug from "debug";
```
tomdale: Can you explain this change? The other imports seem to use single quotes.

packages/@glimmer/compiler-delegates/src/module-unification/compiler-delegate.ts
```diff
@@ -73,7 +73,7 @@ export default class MUCompilerDelegate implements AppCompilerDelegate<TemplateM
     let relativePath = module.replace(/^\.\//, '');
 
     let meta;
-    if (this._builtins[name]) {
+    if ((this as any)._builtins[name]) {
```
tomdale: The type of `this` should be correctly inferred because it's a class method. Do you know why you need this cast?

locks:
Apologies for letting the PR linger. I am closing it due to the amount of conflicts with the current master.

PR #125, opened 2018-05-28
Upgrade to Glimmer VM 0.34.7

tomdale:
This PR updates Glimmer.js to use version 0.34.7 of the Glimmer VM dependencies.

The only notable changes internally are:

- Reverted the monkeypatching of `scheduleRerender` in the test application and replaced it with the `didRender` helper.
- Made user helper references cached to workaround a change in the VM causing helper functions to be called 3x per helper invocation in the template.
- Adapting to changes to the VM's public API mostly around template compilation.

As far as I know there is no impact to Glimmer.js users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
     "packages/@glimmer/*"
   ],
   "devDependencies": {
-    "@glimmer/compiler": "^0.31.0",
     "@types/qunit": "^2.0.31",
     "amd-name-resolver": "^1.0.0",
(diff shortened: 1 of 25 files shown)


tomdale:
@rwjblue Seems to work with the Glimmer Playground in my tests, although npm link is so broken it's hard to know for sure.

rwjblue:
Indeed, thank you for testing.

Issue #132, opened 2018-06-19
Named arguments yielded to handlebars are lost


happycollision:
Sometime between November 11th 2017 and now, [the example](http://tinyurl.com/y8kllzd2) linked from issue #79 stopped working.

The example shows a way of yielding a hash to the calling component. Similar to this:

```handlebars
<SomeThing as |meta|>
  {{logger meta}}
</SomeThing>
```

In the actual example, a component is yielded, but here just imagine a POJO with an `id` property is yielded.

If you were to create a helper (like the example above) that logs to the console, you'd get the following:

```
[{id: 43}]
```

But as soon as you try to use the actual `id` property...

```handlebars
<SomeThing as |meta|>
  {{logger meta.id}}
</SomeThing>
```

 you'd see this in the console:
```
[HelperReference]
```

And if you simply used it in you template without logging (`{{ meta.id }}`), you'd get something like this:
```html
<div class="some-thing">
  [object Object]
</div>
```

It works just fine on the currently running playground, but the same code is broken in newly created Glimmer projects. I have tried to hunt down the issue, but all the packages and dependencies are throwing me for an absolute loop.

FYI: 
```
tests $ ember -v
ember-cli: 3.1.4
node: 9.11.1
os: darwin x64
```

happycollision:
Example Repo: <https://github.com/happycollision/Glimmerjs-Issue-132>

lifeart:
@happycollision can you rice PR into glimmer repo with this failing test?

happycollision:
It seems as though this issue is fixed on master. I suspect (though have not tested) that it is either 65641bffe84ad6ed10428401d3ae0de48c53546d or 0e5bb2cdc661e210e3620fb237be48448107dcd0 which fixes the issue.

PR #133, opened 2018-06-21
Adds regression test for issue #132 which appears fixed on master

happycollision:
As per @lifeart's request, I am adding this test for a broken syntax in v0.9.1 (It is now fixed in `master`). Referenced in issue #132.

I do not know if this is the best place for this test, so suggestions are welcome.

(I did check the test at the v0.9.1 tag, and it fails there.)


rwjblue:
Thank you!!

PR #134, opened 2018-06-25
await for rerender to finish before changing rendering flag

lifeart:
https://github.com/lifeart/glimmer.js/blob/bdd06581cc2717f82d1b94bdae3ee9e510c6be7a/packages/%40glimmer/application/src/application.ts#L334

* _rerender is async function, we need to wait for it


lifeart:
@rwjblue plz

lifeart:
@tomdale

rwjblue:
Thank you! Sorry for the delay in review!

Issue #139, opened 2018-09-13
Remove Dependent Keys On `@tracked`


chadhietala:
As of https://github.com/glimmerjs/glimmer.js/pull/115 `@tracked` no longer requires the dependent keys. We need to do the following to unify this implementation with the one that Ember is going to expose. Below is a list of things we should do to align them:

- [x] Add a deprecation for when a dependent key is passed.
- [x] Ship version 0.11.0 with the deprecation
- [x] Remove the DK deprecation and all the associated code
- [x] Ship version 0.12.0 with the DK code gone

PR #141, opened 2018-09-17
Add deprecation warning for tracked property dependent keys

smfoote:
Step 1 of #139

### DIFF
--- a/packages/@glimmer/component/src/tracked.ts
+++ b/packages/@glimmer/component/src/tracked.ts
@@ -89,6 +89,9 @@ export function tracked(...dependencies: any[]): any {
 
   if (typeof target === "string") {
     return function(target: any, key: string | Symbol, descriptor: PropertyDescriptor) {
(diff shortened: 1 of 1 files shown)


packages/@glimmer/component/src/tracked.ts
```diff
@@ -89,6 +89,9 @@ export function tracked(...dependencies: any[]): any {
 
   if (typeof target === "string") {
     return function(target: any, key: string | Symbol, descriptor: PropertyDescriptor) {
+      if (dependencies.length) {
+        console.warn('The use of dependent keys in @tracked is deprecated (e.g. `@tracked(\'myDependendentKey\')`) and will be removed in v0.12.0');
```
chadhietala: Can you add:

```js
Please remove the "${dependencies.length > 1 ? dependencies.join(', ') : dependencies[0]}" key${dependencies.length > 1 ? 's' : ''}.
```

PR #143, opened 2018-09-20
Remove dependent keys

smfoote:
Last step of #139

### DIFF
--- a/packages/@glimmer/component/src/tracked.ts
+++ b/packages/@glimmer/component/src/tracked.ts
@@ -106,6 +106,7 @@ let CURRENT_TRACKER: Option<Tracker> = null;
 function descriptorForTrackedComputedProperty(target: any, key: any, descriptor: PropertyDescriptor): PropertyDescriptor {
   let meta = metaFor(target);
   meta.trackedProperties[key] = true;
+  meta.trackedComputedProperties[key] = true;
 
   let get = descriptor.get as Function;
   let set = descriptor.set as Function;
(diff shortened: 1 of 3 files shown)


packages/@glimmer/component/src/tracked.ts
```diff
@@ -78,29 +75,16 @@ class Tracker {
...
-  if (typeof target === "string") {
-    return function(target: any, key: string | Symbol, descriptor: PropertyDescriptor) {
-      if (dependencies.length) {
-        console.warn('The use of dependent keys in @tracked is deprecated (e.g. `@tracked(\'myDependendentKey\')`) and will be removed in v0.12.0');
-        console.warn(`Please remove the "${dependencies.length > 1 ? dependencies.join(', ') : dependencies[0]}" key${dependencies.length > 1 ? 's' : ''}`);
-      }
-      return descriptorForTrackedComputedProperty(target, key, descriptor, dependencies);
-    };
+export function tracked(...args: any[]): any {
+  let [target, key, descriptor] = args;
```
chadhietala: Can we add an error somewhere in here to make sure that the usage of `@tracked` is correct? For example:

```
class Foo {
  @tracked() bar = 'foo';
  @tracked('fiz') get bam() {
    return this.fiz + 1;
  }
}
```

These 2 usages are going to result in broken behavior and we should be able to detect this.

packages/@glimmer/component/src/tracked.ts
```diff
@@ -283,19 +269,8 @@ export default class Meta {
...
-  if (dependencies && dependencies.length) {
-    for (let i = 0; i < dependencies.length; i++) {
-      tags.push(meta.tagFor(dependencies[i]));
-    }
-  }
-
-  // Return a combinator across the CP's tags and its dependencies' tags.
-  return combine(tags);
+function combinatorForComputedProperties(meta: Meta, key: Key): Tag {
+  return combine([meta.updatableTagFor(key)]);
```
chadhietala: I'm not sure we even need to call `combine` here. If there is only 1 item in the array there isn't anything to combine.

packages/@glimmer/component/src/tracked.ts
```diff
@@ -283,19 +269,8 @@ export default class Meta {
...
-  // Next, add in all of the tags for its dependencies.
-  if (dependencies && dependencies.length) {
-    for (let i = 0; i < dependencies.length; i++) {
-      tags.push(meta.tagFor(dependencies[i]));
-    }
-  }
-
-  // Return a combinator across the CP's tags and its dependencies' tags.
-  return combine(tags);
+function combinatorForComputedProperties(meta: Meta, key: Key): Tag {
```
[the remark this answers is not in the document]
smfoote: We do need to at least call `meta.updatableTagFor` in the one place that this function is being used. This was the reason my tests were failing.

Issue #146, opened 2018-09-25
Owner Entanglement


chadhietala:
Currently `getOwner` and `setOwner` are used within the component manager. This is an issue because this creates a coupling between the glimmer component manager and `@glimmer/di`. To allow for "Glimmer Components in Ember" we need to remove this entanglement and just have the environment define what it means to set/get the owner.

- [ ] move `getOwner` to the environment
- [ ] move `setOwner` to the environment

locks:
Related PR has been merged.

PR #149, opened 2018-10-02
Move `getOwner` out of component manager

smfoote:
Addresses the first part of #146


packages/@glimmer/application/src/environment.ts
```diff
@@ -77,4 +78,12 @@ export default class Environment extends GlimmerEnvironment {
 
     return new Iterable(ref, keyFor);
   }
+
+  getOwner(): Owner {
+    return getOwner(this);
+  }
+
+  setOwner(obj: Object, owner: Owner): void {
+    return setOwner(obj, owner);
```
chadhietala: No need to return here.

PR #150, opened 2018-10-14
use walk-sync type definitions

givanse:


tsconfig.json
```diff
@@ -38,6 +38,7 @@
     "packages/@glimmer/*/src/**/*.ts",
     "packages/@glimmer/*/index.ts",
     "packages/@glimmer/*/test/**/*.ts",
-    "packages/rollup-plugin-typescript2/src/**/*.ts"
+    "packages/rollup-plugin-typescript2/src/**/*.ts",
+    "packages/@glimmer/app-compiler/walk-sync/index.d.ts"
```
givanse: The latest version of `walk-sync` doesn't get hoisted to the root `node_modules`. About 10 other modules depend on a older version and that is the one that gets hoisted (`0.3.2`). Once more dependencies are upgraded, `walk-sync 0.3.3` will be hoisted and this line could be changed to `packages/walk-sync/index.d.ts`. I guess that would be more correct, but I don't foresee any real downside if this config line was to stay in place.

PR #151, opened 2018-10-16
Create a new class for SSR Applications

chiragpat:
Now that glimmer-vm has a new renderComponent API, we can use that to setup a new ssr application for glimmer with a renderToString API.


packages/@glimmer/ssr/src/application.ts
```diff
@@ -0,0 +1,89 @@
...
+      initialize(registry) {
+        registry.register(`component-manager:/${rootName}/component-managers/main`, ComponentManager);
+      }
+    });
+
+    // Setup registry and DI
+    this.initialize();
+  }
+
+  async renderToStream(componentName: string, data: Opaque, stream: NodeJS.WritableStream) {
```
chadhietala: Just to confirm, this currently isn't piping into a stream as it renders?

chiragpat: Correct, it is not currently piping into a stream. I added it to ensure that the API is future proof for whenever we have support for incremental streaming esque renders.

packages/@glimmer/application/src/application.ts
```diff
@@ -129,22 +148,122 @@ export type Notifier = [() => void, (err: Error) => void];
...
- * The central control point for starting and running Glimmer components.
+ * A Base Application Class that can be shared across different glimmer application implementations.
+ * It sets up a dependency injection and sets up the necessary arguments for any glimmer application.
+ * For example this base class can be reused across a server side glimmer application and a client side glimmer application.
  *
- * @public
+ * @internal
  */
-export default class Application implements Owner {
+export class BaseApplication implements Owner {
```
chadhietala: I think I might be a little bit confused. Don't we need a "base" object that is responsible for setting up the initial registry then effectively becoming a factory that produces clean Application instances that reference the registry? I would assume that we don't want to do all registry setup per request in the SSR case.

chiragpat: The idea behind the BaseApplication was exactly this. The way it is setup right now the SSRApplication instance can be reused across requests and it does not setup on a new registry per request. Specifically I mean the renderToStream and renderToString method can be safely called on the same application instance across different requests.

packages/@glimmer/application/src/application.ts
```diff
@@ -129,22 +150,122 @@ export type Notifier = [() => void, (err: Error) => void];
...
- * The central control point for starting and running Glimmer components.
+ * A base application class that can be shared across different Glimmer application implementations.
+ * It sets up dependency injection and other configuration.
+ * For example, this base class can be reused across a server-side Glimmer application and a client-side Glimmer application.
  *
- * @public
+ * @internal
  */
-export default class Application implements Owner {
+export abstract class BaseApplication implements Owner {
```
tomdale: Can we break this out into a separate file? We generally try to have a single class (as the default export) per file in this codebase.

packages/@glimmer/application/src/environment.ts
```diff
@@ -26,6 +26,28 @@ export interface EnvironmentOptions {
   appendOperations?: DOMTreeConstruction;
 }
 
+export function iterableFor(ref: Reference<Opaque>, keyPath: string): OpaqueIterable {
```
tomdale: Seems like this should be in `iterable.ts` rather than `environment.ts`?

packages/@glimmer/application/src/loaders/bytecode/resolver.ts
```diff
@@ -56,6 +56,7 @@ export default class BytecodeResolver implements RuntimeResolver<TemplateMeta> {
   lookupComponentDefinition(name: string, referrer: TemplateMeta): ComponentDefinition {
     let owner = this.owner;
     let manager = this.managerFor();
+    referrer = referrer || { specifier: null };
```
tomdale: Will this produce a confusing error if you call `renderToString` with an unknown component name? E.g. "invoked using the {{component}} helper in null".

tomdale: We may want to enforce as an invariant that `referrer` can never be null. Are there other cases I'm forgetting?

packages/@glimmer/application/src/loaders/runtime-compiler/loader.ts
```diff
@@ -52,4 +52,8 @@ export default class RuntimeCompilerLoader implements Loader {
       mainLayout.asLayout().compile()
     ));
   }
+
+  getComponentTemplateIterator(app: BaseApplication, env: Environment, builder: ElementBuilder, componentName: string, args: RenderComponentArgs): Promise<TemplateIterator> {
+    throw new Error("Method not implemented.");
```
tomdale: What's needed to support this?

packages/@glimmer/ssr/src/application.ts
```diff
@@ -0,0 +1,89 @@
...
+      await this.renderer.render(templateIterator);
+      env.commit();
+      stream.write(this.serializer.serializeChildren(doc.body));
+      stream.end();
+    } catch (err) {
+      stream.emit('error', err);
+    }
+  }
+
+  async renderToString(componentName: string, data: Opaque): Promise<string> {
```
tomdale: I don't think `Opaque` is exactly the right type here. This type signature allows me to pass `true` or `1234` without type errors, even though I'd get a runtime error from the `convertOpaqueToReferenceDict` function. I think it's more like `Dict<Opaque>`?

packages/@glimmer/ssr/src/application.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    this.serializer = new HTMLSerializer(voidMap);
+    this.registerInitializer({
+      initialize(registry) {
+        registry.register(`component-manager:/${rootName}/component-managers/main`, ComponentManager);
+      }
+    });
+
+    // Setup registry and DI
+    this.initialize();
```
tomdale: Maybe I missed it, but don't we need to create/destroy the container per request? Having a shared container makes it super easy to leak PII between requests.

packages/@glimmer/application/src/application.ts
chiragpat: Yup moved it out and all shared types used across both apps into base app as well.

packages/@glimmer/application/src/environment.ts
chiragpat: Moved as well.

packages/@glimmer/ssr/src/application.ts
chiragpat: Good catch. I completely forgot Opaque does not imply Dict.

chiragpat: Fixed in latest commit.

packages/@glimmer/application/src/loaders/runtime-compiler/loader.ts
chiragpat: I unfortunately haven't had time to dig through all the details needed to implement this here. I was planning on addressing this as a separate issue once this PR was merged.

packages/@glimmer/application/src/loaders/bytecode/resolver.ts
tomdale: Ahh, apologies. Forgot about that case.

tomdale:
Finally got around to merging this! There are some type errors caused by the latest Glimmer VM/SimpleDOM update that I've put some kludge fixes in for. I have some PRs out to SimpleDOM that will allow us to update Glimmer VM to the latest version, then will cut a new release of the VM that includes the fixed types.

PR #157, opened 2018-11-09
Addon blueprint

rondale-sc:
This creates a glimmer addon when passed as the blueprint option to `ember new`

`ember new -b @glimmer/addon-blueprint my-addon` 

It creates a sample helper to assist new addon authors learning.


packages/@glimmer/addon-blueprint/files/ember-addon-index.js
```diff
@@ -0,0 +1,26 @@
...
+const Funnel = require('broccoli-funnel');
+const BroccoliDebug = require('broccoli-debug');
+const MergeTrees = require('broccoli-merge-trees');
+
+module.exports = {
+  name: '<%= name %>',
+  blueprintsPath() {
+    return null;
+  },
+  preprocessTree(type, tree) {
```
rwjblue: I'm not sure if we should make this in the default blueprint. This style worked well for your polyfill addon because we ultimately expect those helpers to be moved into "core", but in general with MU we don't want folks mutating the source apps `src` 🤔

I think the better thing to do (until we have MU support for `{{use` or the like) is generate the files into the host. 

What do you think?

packages/@glimmer/addon-blueprint/files/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
rwjblue: This should probably be fleshed out more, whats in the app `.gitignore`?

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "<%= name %>",
+  "version": "1.1.1",
```
rwjblue: Probably default to either `0.0.0` or `1.0.0`, `1.1.1` feels odd....

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "keywords": [
+    "ember-addon"
+  ],
+  "description": "",
+  "main": "src/index.js",
+  "license": "MIT",
+  "dependencies": {
+    "babel-plugin-external-helpers": "^6.22.0",
```
rwjblue: Sucks that this is required, have we dug into how to fix?

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "broccoli-debug": "^0.6.5",
```
rwjblue: Probably not needed by default

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "broccoli-funnel": "^2.0.1",
```
rwjblue: Probably not needed by default

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "broccoli-merge-trees": "^3.0.1"
```
rwjblue: Probably not needed by default

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "scripts": {
+    "test": "ember test"
```
rwjblue: Should add `lint:ts` / `lint:hbs` at least

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "devDependencies": {
+    "@glimmer/application-pipeline": "^0.11.2",
+    "@glimmer/build": "^0.9.1",
+    "@glimmer/component": "^0.9.1",
+    "@glimmer/inline-precompile": "^1.0.1",
+    "@glimmer/test-helpers": "^0.31.2",
+    "@types/qunit": "^2.0.31",
+    "broccoli-debug": "^0.6.5",
+    "broccoli-file-creator": "^2.1.1",
```
rwjblue: probably not needed

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "configPath": "tests/dummy/config"
+  },
+  "devDependencies": {
+    "@glimmer/application-pipeline": "^0.11.2",
+    "@glimmer/build": "^0.9.1",
+    "@glimmer/component": "^0.9.1",
+    "@glimmer/inline-precompile": "^1.0.1",
+    "@glimmer/test-helpers": "^0.31.2",
+    "@types/qunit": "^2.0.31",
+    "broccoli-debug": "^0.6.5",
```
rwjblue: probably not needed

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "broccoli-file-creator": "^2.1.1",
+    "broccoli-funnel": "^2.0.1",
```
rwjblue: probably not needed

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "broccoli-merge-trees": "^2.0.0",
```
rwjblue: probably not needed

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "ember-cli": "~3.4.3",
```
rwjblue: We can't update to 3.5 due to Broccoli@2 stuffs?

packages/@glimmer/addon-blueprint/files/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "typescript": "~2.6.0"
+  },
+  "resolutions": {
+    "@glimmer/application-pipeline/**/heimdalljs": "^0.2.6"
```
rwjblue: Probably shouldn't need this, right?

packages/@glimmer/addon-blueprint/files/testem.json
```diff
@@ -0,0 +1,15 @@
+{
```
rwjblue: Lets use `testem.js` instead

rwjblue: Hmm, the files from here down _basically_ see like the app files without lots of modifications other than the paths. Can we avoid duplicating them like this?

ember-cli itself avoids duplicating by using a `fileMap` to move the app files into `tests/dummy`

https://github.com/ember-cli/ember-cli/blob/6e84ea4da023960f89404a781375223e6c833609/blueprints/addon/index.js#L151-L176

rondale-sc:
Will do.  I'll have time either during the weekend or Monday to attempt to resolve these points.  Thanks for reviewing!

packages/@glimmer/addon-blueprint/files/ember-addon-index.js
rondale-sc: I'm not sure.  I think the current `use` RFC isn't in a place where we can really use it.  And there are still questions about it.  

I think, so long as we have an ability to automatically move people in a different direction, we'd prefer this approach to polutting consuming applications with a million files.

Maybe making `use` work in block form (which as we discussed might be possible in user space) would be a temporary hack that'd still give us namespace.

rwjblue: I definitely understand your point, but I am still uncomfortable shipping an official addon blueprint that mutates the host app's `src`. 🤔

pzuraq:
Closing this since it's now outdated. Glimmer 2.0 applications will use standard Node bundlers and not rely on the addon build system. We could add a "library" blueprint that creates a simple setup for developing a library though.

Anyways, sorry it didn't make it in, thanks for your efforts here!

PR #159, opened 2018-11-28
Friendly error message when component cannot be found in bytecote meta map.

### LINKED ISSUE #154: Improve error message in BytecodeResolver when templateSpecifier not found in meta map
I bumped into a problem that was caused by a misconfiguration in a library name between the built artifact and a value used at runtime. I ended up getting a message `TypeError: Cannot destructure property 'v' of 'undefined' or 'null'.`

https://github.com/glimmerjs/glimmer.js/blob/763989f920d03c3729b0fd0ab488e911c26abe38/packages/%40glimmer/application/src/loaders/bytecode/resolver.ts#L67

The root problem would have been clearer if the trimmed templateSpecifier and the keys available in the meta map were logged in an error message.

chiragpat:
Fixes #154

### DIFF
--- a/packages/@glimmer/application/src/loaders/bytecode/resolver.ts
+++ b/packages/@glimmer/application/src/loaders/bytecode/resolver.ts
@@ -65,6 +65,11 @@ export default class BytecodeResolver implements RuntimeResolver<TemplateMeta> {
     }
 
     let trimmed = templateSpecifier.replace(this.prefix, '');
+
+    if (!this.meta[trimmed]) {
+      throw new Error(`Could not find component <${trimmed}> invoked from the <${referrer.specifier}> component. Available components are: ${Object.keys(this.meta)}`);
+    }
+
     let { v: vmHandle, h: handle, table: symbolTable } = this.meta[trimmed];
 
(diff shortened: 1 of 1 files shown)


packages/@glimmer/application/src/loaders/bytecode/resolver.ts
```diff
@@ -65,6 +65,11 @@ export default class BytecodeResolver implements RuntimeResolver<TemplateMeta> {
     }
 
     let trimmed = templateSpecifier.replace(this.prefix, '');
+
+    if (!this.meta[trimmed]) {
+      throw new Error(`Could not find ${trimmed} in metadata map for bytecode. Keys that exist in the map: ${Object.keys(this.meta)}`);
```
tomdale: Awesome work! I'm wondering if this error message is too specific to the internal implementation.

Maybe something like `Could not find component <${componentName}> invoked from the <${referrerComponentName>. Available components are: ${Object.keys(this.meta)}`? I think this might better match the mental model of the person who encounters the error.

chiragpat: Yup makes more sense. Fixed it up.

PR #163, opened 2018-12-04
Fixed ESLint issue with component-test blueprint

bhullarg:
There is ESLint issue with files generated by @glimmer/blueprint. "component-test.ts" generated by this blueprint uses "this.render". I have changed "this.render" to "render", which is also used in https://github.com/glimmerjs/glimmer-application-pipeline/blob/master/blueprints/glimmer-component-test/files/src/ui/components/__name__/component-test.ts


tomdale:
Thank you!

PR #168, opened 2019-01-29
[FEAT] Implement Glimmer Components RFC

pzuraq:
This PR turns the `@glimmer/component` package into an Ember addon, allowing it to work in both Ember.js and Glimmer.js applications. This works by keeping the `GlimmerComponent` class definition in a separate file that then gets imported by the Glimmer.js implementation in `src`, and is included in the Ember.js implementation via `addon/index.js`.

Notes:

* Had to hack around `@glimmer/di`, it wants to be able to inject things universally but that is not allowed in the final design for `GlimmerComponent`. When we receive a `ComponentFactory`, we pull the class definition off it and use it directly instead. 
* The build has been simplified so all additional Ember Addon files will be pushed to the final dist. This means `@glimmer/blueprint` doesn't need to be treated too specially anymore, it just gets ignored by the packaging step so we don't build an empty `dist`.
* I have Prettier on, so some files have been made ✨more beautiful✨. We should add it to the project as a whole.


pzuraq:
Tested in both a Glimmer.js and Ember.js app generated using the default blueprints. The Glimmer.js app was hard to test, due to the way the application pipeline works for it. I ended up having to actually copy/paste the dist into `node_modules` but that seemed to work, if anyone knows a better way I'm all ears 😄

build/broccoli/optimize.js
```diff
@@ -12,17 +10,22 @@ module.exports = function(jsTree) {
...
   let plugins = [
-    [debugMacros, {
-      envFlags: {
-        source: '@glimmer/env',
-        flags: {
-          DEBUG: !isProduction,
-          CI: !!process.env.CI
-        }
-      },
+    [require.resolve('babel-plugin-debug-macros'), {
```
tomdale: Were the changes in this file needed for Ember compatibility, or just general cleanup?

pzuraq: Something happened as I was upgrading that caused the `babel-plugin-debug-macros options have changed` deprecation message to start popping up everywhere, so I figured it would be good to clean it up and update.

PR #175, opened 2019-02-15
Re-enable lifecycle hooks for Glimmer.js environments

tomdale:
This commit restores the `didInsertElement` and `didUpdate` lifecycle hooks and `element`/`bounds` properties that were removed during the work to enable @glimmer/component to work as an Ember addon.

Our intent is for developers to use an element modifier-based API to access the DOM, rather than component hooks and properties. This is supported in the Ember environment today. However, the required element modifier manager API is not yet implemented in Glimmer.js.

Not having this capability blocks many Glimmer.js apps from upgrading, so this commit restores that functionality and allows those apps to upgrade to 0.14. Once we implement element modifier manager support in Glimmer.js, we can re-remove these APIs and align the Glimmer/Ember environments.

### DIFF
--- a/packages/@glimmer/component/src/bounds.ts
+++ b/packages/@glimmer/component/src/bounds.ts
@@ -1,0 +1,22 @@
+import { Bounds as VMBounds } from '@glimmer/interfaces';
+
+/**
+ * Contains the first and last DOM nodes in a component's rendered
+ * template. These nodes can be used to traverse the section of DOM
+ * that belongs to a particular component.
+ *
+ * Note that these nodes *can* change over the lifetime of a component
+ * if the beginning or ending of the template is dynamic.
+ */
+export default class Bounds {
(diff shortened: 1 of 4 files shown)


rwjblue:
I understand the goal here, but I'd rather work to fix the modifier situation 🤔.

FWIW, we will need _another_ breaking change release when we revert this revert...

tomdale:
Discussed this a bit with @rwjblue offline. I'm going to merge this PR and cut a new alpha release, so I can land the upgrade work I've been doing in the Glimmer Playground app. However, I agree that we should not create churn here, and Ember/Glimmer interop is our top priority.

I wasn't sure when we might have time to work on adding support for modifier managers to Glimmer, but @rwjblue thinks he can work on it next week. We will hold off cutting a new stable release if we can get that in soon.

rwjblue:
Looks like it will need a rebase (due to prettier changes).

PR #177, opened 2019-02-15
Fix TS strict mode errors in @glimmer/application

tomdale:
Fixes strict mode errors in TypeScript in `@glimmer/application`. This doesn't enable strict mode in `tsconfig.json` yet because some of the other packages have strict mode errors.


mike-north:
types LGTM

rwjblue:
Thanks for reviewing @mike-north!

PR #180, opened 2019-03-16
[BUGFIX] Fixes args tracking in Glimmer component

pzuraq:
`@tracked` is not yet avalibale for general usage in Ember apps,
so we were relying on using the interop with `get` and
`set` to track args. However, we never actually used `get`,
so it wasn't picking it up. This ensures that all gets/sets of
the field go through the proper codepaths, and will be autotracked.


rwjblue:
This seems fine for now, but I would like to update to something akin to what was done in:

https://github.com/glimmerjs/glimmer.js/commit/a7d5fb64d3a7691dbc5a7b4caebd13f167754d33

Even though that was for glimmer.js's version, the code is roughly the same in Ember (going back a ways now).

PR #182, opened 2019-03-25
[@glimmer/component] Make component generic over args

mike-north:
This allows TS consumers to (optionally) benefit from formalizing the interface a component has with the outside world.

```ts
import Component from '@glimmer/component';

export default class Foo extends Component<{ word: string}> {

  printWord() {
     console.log(this.args.word); // type-safe access to args
  }
}

```

### DIFF
--- a/packages/@glimmer/component/src/component.ts
+++ b/packages/@glimmer/component/src/component.ts
@@ -9,7 +9,7 @@ export interface Bounds {
   lastNode: Node;
 }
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Change looks good to me, and mirrors what was done in the types for the "ember consumers" 

https://github.com/glimmerjs/glimmer.js/blob/b9a016573682266b59c3ab75c24ec22f8ea5fbc0/packages/%40glimmer/component/addon/index.ts#L9

PR #183, opened 2019-03-28
Align yarn and npm install commands

bartocc:
I believe the `yarn add` and `npm install` commands were not equivalent.

The former installed in `dependencies` while the later installed in `devDependencies.`

This changes makes sure both examples are the same and install in `devDependencies.`


locks:
Thanks!

PR #185, opened 2019-04-16
Bump glimmer-vm to 0.40.1

chiragpat:
- Renaming nextSibling parameter in `in-element` helper to insertBefore


tomdale:
Looks good, thank you!

PR #191, opened 2019-04-30
Adding support to pass in a dynamic scope into the ssr glimmer app

chiragpat:


tomdale:
For some context here: we want to support a low-level primitive for providing the ambient dynamic scope when rendering a root component. This has been supported in Glimmer VM for a long time, but not exposed in Glimmer.js.

This is not public, stable API we want to commit to, but we want to unblock people who need this capability and are willing to opt-in to an unstable API. We can make our intentions clearer by making the dynamic scope an option whose key is a symbol that must be imported and has a sufficiently descriptive name.

@chiragpat I'm not sure about the `SET_` in `SET_INTERNAL_DYNAMIC_SCOPE`, since options are typically nouns, not verbs. I would be comfortable with `INTERNAL_DYNAMIC_SCOPE`, but open to bikeshedding on this.

packages/@glimmer/application-test-helpers/src/app-builder.ts
```diff
@@ -18,9 +18,10 @@ import { buildAction, mainTemplate } from '@glimmer/application';
 import { compilable } from '@glimmer/opcode-compiler';
 import { Metadata } from '../../application/src/loaders/bytecode/loader';
 import { SimpleDocument } from '@simple-dom/interface';
-import { SSRApplication } from '@glimmer/ssr';
+import { SSRApplication, SET_INTERNAL_DYNAMIC_SCOPE } from '@glimmer/ssr';
```
tomdale: Given this is functionality we will probably want to expose for client-side rendering use cases as well, I'm thinking this symbol should be in `@glimmer/application` so it can be shared across all users of `BaseApplication`?

chiragpat: Ah good point. I will move it over.

chiragpat:
> @chiragpat I'm not sure about the `SET_` in `SET_INTERNAL_DYNAMIC_SCOPE`, since options are typically nouns, not verbs. I would be comfortable with `INTERNAL_DYNAMIC_SCOPE`, but open to bikeshedding on this.

Agreed better key name will go ahead and update.

chiragpat:
> For some context here: we want to support a low-level primitive for providing the ambient dynamic scope when rendering a root component. This has been supported in Glimmer VM for a long time, but not exposed in Glimmer.js.
> 
> This is not public, stable API we want to commit to, but we want to unblock people who need this capability and are willing to opt-in to an unstable API. We can make our intentions clearer by making the dynamic scope an option whose key is a symbol that must be imported and has a sufficiently descriptive name.

I will add a comment on top of the symbol declaration explaining this as well.

PR #193, opened 2019-05-02
Add JIT mode support for renderComponent

chiragpat:
- Fixing regression with SSR environment
- Adding more tests to prevent future regressions


tomdale:
This is a nice cleanup, thank you!

PR #201, opened 2019-07-19
[BUGFIX] Updates the assignment of args to reflect the arg proxy

pzuraq:
The refactor to use a proxy for tracking individual updates to args
requires us to make some changes here in how we assign arguments. Also
adds the `updateHook` option to capabilities. This solution should be
compatible with Ember pre and post enabling the tracked canary flag,
but can be made better with ember-compatibility-helpers once we have
a version of Ember to target.


rwjblue:
> can be made better with ember-compatibility-helpers once we have a version of Ember to target

You should be able to use `gte('3.13.0-alpha.0')` already.

rwjblue:
Looks like the CI failure is a bunch of typescript errors when ran against canary builds of ember-source.

rwjblue:
@tomdale - Isn't this canary failure the same as you hit over in #202? I thought we chatted about it and that you had path forward, but I can't recall what it was. Do you recall?

pzuraq:
@rwjblue I wanted to be a bit more sure before we pinned it here, it'll be annoying if we release a version and it breaks later, but I can also add conditional code if you prefer. Definitely can put the test in the conditional at least.

As for the test failures, there's a legitimate one in there, it's caused by the fact that CPs can't work with the args proxy yet. Fix is here: https://github.com/emberjs/ember.js/pull/18217

packages/@glimmer/component/test/ember/integration/components/glimmer-component-test.js
```diff
@@ -1,10 +1,12 @@
 import GlimmerComponent from '@glimmer/component';
-import { module, test } from 'qunit';
+import { module, test, skip } from 'qunit';
```
rwjblue: We must not be linting something here, `skip` is unused (and that lint failure isn't failing the build) 🤔 

Not a giant issue, just isn't used any longer

tzellman:
Great work here!

There is currently an odd proxy-related error that occurs on ember canary (or 3.13.x), and a few of us believe that this will fix the error. See this example repo which highlights the problem: https://github.com/justinross/e-octane-bug

tzellman:
Thank you for the quick response @rwjblue and for the changes @pzuraq!

🎉

vitch:
Thanks for this fix!

We had the same problem demonstrated in the repo above:

> TypeError: 'ownKeys' on proxy: trap returned extra keys but proxy target is non-extensible

And [as discussed here](https://github.com/justinross/e-octane-bug/issues/1) updating `@glimmer/component` to `0.14.0-alpha.11` *and* updating `ember-source` to `https://s3.amazonaws.com/builds.emberjs.com/canary/shas/e8a6420f39d0a6a6f67a85f62f30fbb5c4339507.tgz` (from `3.13.0-beta.2`) fixed the issue.

I'm wondering what it is about the canary that makes it work. Are there changes in `ember-source` or features that need to be enabled for this to work?

pzuraq:
There were changes in `ember-source` yes. They are being backported to the next beta cycle.

vitch:
It looks like the relevant fix in `ember-source` regressed in [this range](https://github.com/emberjs/ember.js/compare/755ea5dbe65d91e0d650707da740aa6900d0a755...eb5226a230b7066608e3cd1c0045917453ec9572). 

*Or* I'm misunderstanding what this was meant to fix. I put together a a reproduction [here](https://github.com/vitch/ember-octane-test) - should that work?

vitch:
Discussed on discord and fixed by @pzuraq in [ember.js#18274](https://github.com/emberjs/ember.js/pull/18274) - thanks!

PR #202, opened 2019-07-22
Add support for Babel legacy decorators

tomdale:
TypeScript and Babel differ slightly in how they transpile decorated class fields with initial values (e.g. `@tracked name = "Chris"`. In TypeScript, class fields do not have a descriptor, and initialization is compiled into an assignment of the `name` property in the constructor. In Babel, class fields _do_ have a descriptor object and initialization is captured as an `initializer` function on that descriptor.

Currently, the implementation of the `tracked` decorator in Glimmer.js only works with TypeScript decorators. This PR updates the implementation to also be compatible with Babel decorators, which expect a descriptor to be returned from decorators on class fields, and is based on the implementation of `tracked` in Ember.js.

To verify compatibility, this PR also upgrades the build pipeline to Babel v7, and updates the test suite to run tracked property tests against classes compiled with both TypeScript and Babel.

**As part of the Babel 7 upgrade, this PR eliminates AMD versions of the Glimmer.js packages from the build.** These were not being used in tests and are not used by any downstream consumers I'm aware of. Because we ship standard JS modules, downstream consumers that need AMD should be able to compile a version themselves fairly easily.


pzuraq:
Overall looks good, aside from test failures

PR #204, opened 2019-08-12
Adding back didUpdate hook to glimmer.js component-manager

chiragpat:
- This is deprecated API
- This will be removed once we have modifier support in glimmer.js

### DIFF
--- a/packages/@glimmer/component/src/component-manager.ts
+++ b/packages/@glimmer/component/src/component-manager.ts
@@ -228,7 +228,15 @@ export default class ComponentManager
 
   didUpdateLayout() {}
 
-  didUpdate() {}
+  didUpdate(bucket: ComponentStateBucket) {
+    if (DEBUG && bucket instanceof TemplateOnlyComponentDebugBucket) {
+      return;
+    }
+    if (!bucket) {
+      return;
+    }
+     bucket.component.didUpdate();
+  }
 
   getDestructor(bucket: ComponentStateBucket): Destroyable {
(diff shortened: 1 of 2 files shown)


tomdale:
For some context: we haven't released a non-alpha version of Glimmer.js in quite awhile, despite a lot of work happening and many alpha releases since 0.13.0.

The reason for this is because, as part of stabilizing Glimmer components for use in Ember applications, we removed lifecycle hooks (like `didUpdate`) and DOM access (like `this.bounds`) in favor of [render modifiers](https://github.com/emberjs/ember-render-modifiers), which have fewer gotchas and are better aligned with the Octane programming model.

However, while render modifiers are available in Ember, we have not yet had the bandwidth to bring them to Glimmer.js. Without render modifiers and without lifecycle hooks and `this.bounds`/`this.element`, we didn't feel comfortable releasing a new Glimmer.js version, because any app that needs to access the DOM (which, in practice, is most of them) wouldn't be able to upgrade.

In lieu of indefinitely delaying a new Glimmer.js release, we will accept a temporary divergence between Glimmer.js and Ember, and, for now, reintroduce lifecycle hooks and DOM access within the Glimmer.js environment. This should allow Glimmer.js applications to upgrade to the latest version without major changes, and puts us on the path of releasing a "1.0" version of the `@glimmer/component` package that can be included by default in Ember without violating Ember's SemVer guarantees.

The plan of record is still to bring render modifiers to Glimmer.js, and in general support writing components that are portable between both environments. We just want to touch back down to reality with a release first before diving into that work.

PR #206, opened 2019-08-13
Disable updateComponent hook in 3.13.0

tomdale:
Ember 3.13 changes how component `args` participate in the autotrack stack. The `args` object is guaranteed to be stable, and each arg is tracked on its own. This means we no longer need to set the `args` property on a component instance to invalidate tracked getters that rely on an arg. 3.13 also implements a new capability flag that disables the `updateComponent` hook altogether.

To preserve backwards compatibility, this PR implements the `updateComponent` hook when consumed from Ember 3.12 and earlier (including setting `args` for every update) but removes it in Ember 3.13.0-beta.1 and later, as well as setting the capability flag to instruct Ember not to call the hook at all.

This should fix the currently failing tests when run against the Ember beta and canary channels.


tomdale:
@chiragpat Oy, looks like I hit undo too many times in my editor. Updating…

PR #209, opened 2019-08-21
Ensure the tags Set is properly converted to an Array

### LINKED ISSUE #208: ES5 transpilation: UpdatableTag.compute errors with nested tracked properties
Hi there!

I am upgrading a Glimmer app from 0.7 to 0.13, and I have come across the following error.  Unfortunately my project requires ES5 support, which is a factor here, as it otherwise works for ES6 compilation.  I have tested against 0.11.2, 0.13, and 0.14-alpha.12 with the same results, and it can be reproduced on a brand new Glimmer app when enabling ES5 output.

The gist: _when compiling to ES5, if a tracked property relies on another tracked property, all rendering and interaction will stop working without any sort of error._

Let me start by saying that if I thought this was some IE11 bug or something that seems not worth trying to fix, I wouldn't file the issue.  However, it seems like a minor thing that can be solved to continue making Glimmer ES5 compatible.

To reproduce:

1. `ember new es5issue -b @glimmer/blueprint`
2. Change the last line of `config/targets.js` to `module.exports = 'es5';`. Also reproducible by adding `IE 11` to the list of browsers.
3. Include `babel-polyfill` for the essentials.  You can `npm install babel-polyfill` and then add the following as the first line to `index.ts`: `import '../../node_modules/babel-polyfill/browser';`.  Also reproducible by just adding a `<src>` tag to `index.html` to pull it in from Babel's CDN.
4. In the main component, use the demo code from the Glimmer Playground as an example:

```
component.ts:

import Component, { tracked } from '@glimmer/component';
export default class extends Component {
  @tracked count = 1;
  increment() {
    this.count++;
  }
}

template.hbs:

<h1>Welcome to the Glimmer Playground!</h1>
<p>You have clicked the button {{count}} times.</p>
<button onclick={{action increment}}>Click</button>
```

If you run this, it works as expected.  However, if we simply add another tracked property that reads `count`, it will fail:

```
component.ts:

...
  @tracked
  get anotherCount() {
    return this.count; 
  }
...

template.hbs:

...
{{anotherCount}}
...
```

This should work, and it does work via ES6, but it fails here... and without any kind of console error or warning about what happened.

Chasing the rabbit down the hole a bit more, I looked through the renderer code and discovered that if I added the following to my `index.ts`, I could find out if the renderer failed:

```javascript
app._notifiers.push([ (() => console.log('success')), ((e) => console.log('error', e))]);
```

By doing this, I found a helpful error message that was getting swallowed:

```
error TypeError: this.tag.value is not a function
    at UpdatableTag.compute (app.js:4215)
    at UpdatableTag.value (app.js:4128)
    at app.js:3994
    at TagWrapper.value (app.js:3981)
```

In this case, it appears it is expecting `this.tag` to be an `UpdatableTag`, but instead it is a `Set` containing a single `TagWrapper` referencing an `UpdatableTag`.  It looks like the ES6 version has a `TagsPair` object which does have a `value` method, but the ES5 version is a `Set` that does not have a `value` method.

Currently, to fix this, I monkey-patched `UpdatableTag` in my project's `index.ts` by checking for a Set and (recursively) finding the maximum value:

```javascript
import { UpdatableTag } from '@glimmer/reference';
...
function tagValue(tag) {
  if (tag.value) {
    return tag.value();
  } else if (tag instanceof Set) {
    let value = -Infinity;
    tag.forEach((item) => {
      const v = tagValue(item);
      if (v > value) {
        value = v;
      }
    });
    return value;
  } else {
    throw new Error('Cannot determine tag value');
  }
}

UpdatableTag.prototype.compute = function() {
  return Math.max(this.lastUpdated, tagValue(this.tag));
};
```

It is not pretty, but it appears to solve the problem and seems to evaluate the same way as `TagsPair`.

christophersansone:
Fixes #208 .

TL;DR The original implementation works fine when targeting modern browsers, but when targeting ES5, the resulting array _contains_ the Set rather than being flattened to contain the _items_ in the set.  `Array.from` seems like the safer, more canonical approach here, since the code is attempting to create a new Array from an array-like object such as a Set.

### DIFF
--- a/packages/@glimmer/tracking/src/tracked.ts
+++ b/packages/@glimmer/tracking/src/tracked.ts
@@ -25,7 +25,7 @@ class Tracker {
     let { tags } = this;
 
     if (tags.size === 0) return CONSTANT_TAG;
-    return combine([...tags]);
+    return combine(Array.from(tags));
   }
 }
 


rwjblue:
Thanks for working through this one!

rwjblue:
Also, just a note for those wondering how `Array.from` is supposed to work in ES5: any apps wanting to use ES5 targets (e.g. IE11) will _already_ be required to be using a polyfill and thus `Array.from` would be present.

PR #214, opened 2019-09-26
Implement Ember-compatible Custom Component Manager API

tomdale:
Ember RFCs 213 and 392 define a public API for registering custom component managers. Component managers allow addons to provide alternative component APIs on top of Glimmer VM. Support for the Glimmer component API (exported from @glimmer/component) in Ember applications is implemented in terms of this public API.

Previously, Glimmer.js itself relied on the private, super-low-level component manager API provided by the Glimmer VM. This meant that we had to maintain two different implementations of the Glimmer component manager: one used by Glimmer.js, the other used by Ember.

With this change, Glimmer.js now implements the same public component manager API as Ember, allowing the component manager implementation to be (mostly) shared across both environments.

Another benefit of this change is that we can significantly reduce the API surface area exported by the @glimmer/component package. This helps reduce the risk of breakage for Ember users using TypeScript, as the @glimmer/component types no longer rely on the voluminous, more complex types of the Glimmer VM packages.


packages/@glimmer/component/src/component-manager.ts
```diff
@@ -1,259 +1,41 @@
...
-      ARGS_SET.add(snapshot);
-    }
-
-    return Object.freeze(snapshot);
-  }
-}
-
-const EMPTY_SELF = new RootReference(null);
+const CAPABILITIES = capabilities('3.13', {
+  asyncLifecycleCallbacks: true,
```
pzuraq: should this have `destructor: true`?

packages/@glimmer/application/src/components/component-managers/custom.ts
```diff
@@ -0,0 +1,365 @@
...
+          `You attempted to set ${definition.ComponentClass.class}#${String(
+            prop
+          )} on a components arguments. Component arguments are immutable and cannot be updated directly, they always represent the values that are passed to your component. If you want to set default values, you should use a getter instead`
+        );
+
+        return false;
+      };
+    }
+
+    let namedArgsProxy = new Proxy({}, handler);
```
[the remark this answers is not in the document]
rwjblue: Glimmer.js applications doesn't support IE11 already. As long as this doesn't affect _Ember_ apps (which do support IE11 still) it should be fine.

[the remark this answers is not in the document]
tomdale: Agreed with @rwjblue. Given this change is localized to code that is not used by the Ember component manager, I think we're in the clear here.

packages/@glimmer/application/src/components/component-managers/custom.ts
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  didCreate({ delegate, component }: CustomComponentState<ComponentInstance>) {
+    if (hasAsyncLifecycleCallbacks(delegate)) {
+      delegate.didCreateComponent(component);
+    }
+  }
+
+  didUpdate({ delegate, component }: CustomComponentState<ComponentInstance>) {
+    if (hasAsyncLifecycleCallbacks(delegate)) {
```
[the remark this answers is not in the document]
tomdale: Good catch! This was just an oversight on my part.

packages/@glimmer/application/src/components/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const MANAGERS: WeakMap<object, ManagerWrapper<unknown>> = new WeakMap();
+
+const getPrototypeOf = Object.getPrototypeOf;
+
+export type ManagerFactory<ManagerDelegate> = (owner: Owner) => ManagerDelegate;
+
+export interface ManagerWrapper<ManagerDelegate> {
+  factory: ManagerFactory<ManagerDelegate>;
+  internal: boolean;
```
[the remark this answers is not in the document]
tomdale: You're right that I was trying to minimize the delta between the implementations. I think you're right that it's more consistent with other changes in this PR to align the overall approach, but not try to preserve line-for-line parity.

packages/@glimmer/application/src/components/component-managers/custom.ts
```diff
@@ -0,0 +1,375 @@
...
+      ComponentClass,
+      delegate,
+      definition: this
+    };
+  }
+}
+
+export interface ComponentFactory<C = {}> {
+  class: C;
+}
```
chancancode: Newline: you really should turn on the vscode setting 😄

packages/@glimmer/component/src/component-manager.ts
tomdale: Yes, good catch. There were no tests for the `willDestroy` hook. I updated the lifecycle hooks ordering test to include destruction.

packages/@glimmer/application/src/components/bounds.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Contains the first and last DOM nodes in a component's rendered
+ * template. These nodes can be used to traverse the section of DOM
+ * that belongs to a particular component.
+ *
+ * Note that these nodes *can* change over the lifetime of a component
+ * if the beginning or ending of the template is dynamic.
+ */
+export default class Bounds {
```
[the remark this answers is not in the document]
tomdale: Yes this is pre-existing behavior, this PR just moves it from `@glimmer/component` to `@glimmer/application`.

packages/@glimmer/application/src/components/component-managers/custom.ts
```diff
@@ -0,0 +1,375 @@
...
+  dynamicTag: false,
+  wrapped: false,
+  prepareArgs: false,
+  createArgs: true,
+  attributeHook: false,
+  elementHook: false,
+  updateHook: true,
+  createCaller: false,
+  dynamicScope: true,
+};
```
chancancode: FWIW, I think it's silly that Ember uses the same capabilities for _all_ custom components. The point of the design in the RFC is to shim the public capabilities into the underlying glimmer ones. I think we can fix it later, that's not a new thing here, but I never had the chance to mention it.

packages/@glimmer/application-test-helpers/src/app-builder.ts
```diff
@@ -242,6 +236,20 @@ export class AppBuilder<T extends TestApplication> {
   }
 }
 
+const CAPABILITIES = {
```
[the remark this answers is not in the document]
tomdale: Rationale?

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  ComponentDefinitionState,
+  AotComponentDefinition,
+  JitComponentDefinition,
+} from '../component-definition';
+
+export const CAPABILITIES: ComponentCapabilities = {
+  attributeHook: true,
+  createArgs: true,
+  createCaller: false,
+  createInstance: true,
```
chancancode: This should be DEBUG only (though I think there may be a VM bug when this is set to false... unsure if that's fixed now).

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  dynamicLayout: false,
+  dynamicScope: false,
+  dynamicTag: true,
```
chancancode: Is this right?

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+
+import { RootReference, TemplateOnlyComponentDebugReference } from '../../references';
+import {
+  ComponentDefinitionState,
+  AotComponentDefinition,
+  JitComponentDefinition,
+} from '../component-definition';
+
+export const CAPABILITIES: ComponentCapabilities = {
+  attributeHook: true,
```
chancancode: Do we need this?

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  createArgs: true,
+  createCaller: false,
+  createInstance: true,
+  dynamicLayout: false,
+  dynamicScope: false,
+  dynamicTag: true,
+  elementHook: true,
```
chancancode: Do we need this?

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  prepareArgs: false,
+  updateHook: true,
```
chancancode: Do we need this?

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  wrapped: false,
+};
+
+export class ComponentStateBucket {
+  public name: string;
+
+  constructor(public args: CapturedArguments) {
+    this.args = args;
```
chancancode: not needed

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+    // get stripped in production mode and no bucket will be instantiated.
+    return DEBUG ? new TemplateOnlyComponentDebugBucket(state.definition) : null;
+  }
+
+  getSelf(bucket: TemplateOnlyComponentDebugBucket): PathReference {
+    return DEBUG ? new TemplateOnlyComponentDebugReference(bucket.definition.state.name) : EMPTY_SELF;
+  }
+
+  getTag(): Tag {
+    return CONSTANT_TAG;
```
chancancode: I think this needs to be args.tag actually

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  didCreateElement() {}
+  didRenderLayout() {}
+  didCreate() {}
+  update() {}
+  didUpdateLayout() {}
+  didUpdate() {}
```
chancancode: Does the VM types still make us implement these when the hooks are off? if so that's too bad but 🆗

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+  getDestructor(): Destroyable {
+    return NOOP_DESTROYABLE;
+  }
```
chancancode: Should be able to just return `null` here?

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+const NOOP_DESTROYABLE = { destroy() {} };
```
chancancode: Not needed

packages/@glimmer/application/src/components/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function getComponentManager<T>(obj: any): undefined | ManagerFactory<ManagerDelegate<T>> {
+  let wrapper = getManager<ManagerDelegate<T>>(obj);
+
+  if (wrapper && wrapper.type === 'component') {
+    return wrapper.factory;
+  } else {
+    return undefined;
+  }
+}
```
chancancode: newline

packages/@glimmer/application/src/loaders/bytecode/resolver.ts
```diff
@@ -172,3 +153,26 @@ export default class BytecodeResolver implements AotRuntimeResolver<TemplateMeta
...
+  const { factory } = getManager(ComponentClass);
+
+  return new CustomComponentDefinition(
+    name,
+    componentFactory,
+    factory(owner) as ManagerDelegate<unknown>,
+    handle,
+    symbolTable
+  );
+}
```
chancancode: newline

packages/@glimmer/component/test/browser/args-test.ts
```diff
@@ -133,115 +58,4 @@ test('Getters that depend on `args` re-render correctly', async function(assert)
...
-    .component('ParentComponent', ParentComponent)
-    .template('ChildComponent', '<div></div>')
-    .component('ChildComponent', ChildComponent)
-    .boot();
-
-  setPropertyDidChange(function() {
-    app.scheduleRerender();
-  });
-});
+});
```
chancancode: newline

packages/@glimmer/component/test/browser/rendering-test.ts
```diff
@@ -15,71 +12,4 @@ test('A component can be rendered in a template', async function(assert) {
...
-    await buildApp()
-      .template('Main', '<div><HelloWorld></HelloWorld></div>')
-      .template('HelloWorld', '<h1>Hello, {{firstName}} {{lastName}}!</h1>')
-      .component('HelloWorld', HelloWorldComponent)
-      .boot();
-
-    update();
-  });
-}
+});
```
chancancode: newline

packages/@glimmer/application/src/components/component-managers/custom.ts
```diff
@@ -0,0 +1,365 @@
...
+    }
+  }
+
+  didRenderLayout({ delegate, component }: CustomComponentState<ComponentInstance>, bounds: VMBounds) {
+    if (hasDidRenderLayout(delegate)) {
+      delegate.__glimmer__didRenderLayout(component, new Bounds(bounds));
+    }
+  }
+
+  didUpdateLayout() {}
```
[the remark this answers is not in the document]
tomdale: I believe it's only used for inspector debugging, not part of the component API.

packages/@glimmer/component/src/component-manager.ts
tomdale: Yes, good catch! The `willDestroy` hook wasn't covered in the tests, so I updated the lifecycle hook ordering test to now cover destruction.

packages/@glimmer/application/src/components/component-managers/custom.ts
```diff
@@ -0,0 +1,365 @@
...
+  extends ManagerDelegate<ComponentInstance> {
+    capabilities: Capabilities & { updateHook: true };
+    updateComponent(instance: ComponentInstance, args: Args): void;
+  }
+
+export function hasUpdateHook<ComponentInstance>(
+  delegate: ManagerDelegate<ComponentInstance>
+): delegate is ManagerDelegateWithUpdateHook<ComponentInstance> {
+  return delegate.capabilities.updateHook;
+}
```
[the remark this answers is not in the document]
tomdale: Added

packages/@glimmer/application/src/components/component-managers/custom.ts
tomdale: Really it should be a lint. 😁 My hard drive crashed and I had not re-enabled it since reinstalling VS Code.

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: No

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: Setting this to `false` seems to cause the VM to generate invalid bytecode.

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: Are you sure? I thought the tag returned from `getTag()` was only used to determine whether the update hooks should run. Given that TOs don't have any lifecycle hooks to call, are there other cases relying on this invalidating when args change?

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: Good call

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: Nice catch, surprised this isn't caught by tslint

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: I was able to remove the no-op `update()` and `didCreateElement()` hooks, but the other hooks are always called and removing them causes a runtime exception.

packages/@glimmer/application-test-helpers/src/app-builder.ts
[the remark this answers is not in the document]
chancancode: TypeScript would have inferred the “wrong” type here: it would think capabilities is a pojo with whatever fields you have here. If you made a typo or missed a field, you will only get the error much later when you try to use this value by passing it to another API, and often a confusing error. If you had said this is `CAPABILITIES: ComponentCapabilities` here typescript can point out any errors or omission inline instead of cascading it down. It also makes future upgrades easier

packages/@glimmer/application/src/components/component-managers/template-only.ts
chancancode: Yeah so I guess the bug is still there. Do you mind filing an issue on the VM repo?

packages/@glimmer/application/src/components/component-managers/template-only.ts
chancancode: Sounds like we need more capabilities. I’m surprised we ended up having public ember capabilities for these things but not in the VM 🤷🏻‍♂️

packages/@glimmer/application-test-helpers/src/app-builder.ts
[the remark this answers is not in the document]
tomdale: I think in practice it would get caught by the `getComponentCapabilities()` hook immediately below, but I have no objections.

Is the rule that we _always_ require annotating types for object literal data structures that cannot be inferred? I don't know if we're consistent about this in Glimmer.js/Glimmer VM. If that's not the rule of thumb, what is?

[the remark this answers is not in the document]
wycats: @tomdale yes, I think that's a good practice.

When you're defining the capabilities list here, you really do want TypeScript to give you an error if you didn't supply a correct capabilities list, right here.

The problem is that object literals never "cannot be inferred". When you write an object literal, it always has the type of whatever you typed, and therefore will always type check.

Another way to think about it is that type inference is useful if you know that a particular type will type check, and you want to avoid typing the type. It's unhelpful to rely on type inference in a situation where the types might be wrong in the first place.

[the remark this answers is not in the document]
chancancode: https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=8&pc=8#code/MYewdgzgLgBAZiEAuGAxRMC8MDe9EoBEARgIYBOhANDGeUWQF6EwC+A3AFCcCWYUAU3JxSwAWgx4EyGNHJ8A5m25wArmGBQe4eAAoA+inQgAlLk6sVu6SfZA

In this playground, if you delete the type for `const foo`, you will stop getting the type error for the extra field. This is what I meant by making upgrades easier: if we removed a capability we want to get an error here.

packages/@glimmer/application/src/components/component-managers/template-only.ts
tomdale: https://github.com/glimmerjs/glimmer.js/issues/215

packages/@glimmer/application-test-helpers/src/app-builder.ts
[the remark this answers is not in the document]
tomdale: @chancancode You wouldn't get an error when passing it as an argument, but you would get an error caused by the unused property when you return from a method whose return type is specified: https://www.typescriptlang.org/play/#code/MYewdgzgLgBAZiEMC8MDe9EC4YCIBGAhgE64A0MRxOBhAXrjAL4DcAUGwJZhQCmxcQsF4wAYonSYQOaMW4BzZhzgBXMMCidwMALaEA1r3EgAFAEocx9Gxi2YxXlBXEwkhNLxVylEjSINmdiZlNQ0tVzgTAH1LRDNrYLZI9zMWIA

But in any case, I agree this gives you better checking, I have no objections, and have updated the code to specify the type explicitly.

packages/@glimmer/application/src/components/component-managers/custom.ts
pzuraq: I was wondering about this! Figured we could come back to it later, for sure, but good to know

packages/@glimmer/application/src/components/component-managers/template-only.ts
```diff
@@ -28,24 +27,23 @@ import {
...
   createInstance: true,
   dynamicLayout: false,
   dynamicScope: false,
-  dynamicTag: true,
-  elementHook: true,
+  dynamicTag: false,
+  elementHook: false,
   prepareArgs: false,
-  updateHook: true,
+  updateHook: false,
```
rwjblue: FWIW, in Ember this (and `createArgs`) are conditionally set to `true` when the render debug support is enabled. See 

https://github.com/emberjs/ember.js/blob/016f8066d0b484f3f69ce8c337be3b8b8aa5fce4/packages/%40ember/-internals/glimmer/lib/component-managers/template-only.ts#L18-L29

This is specifically, so that we can properly update the render tree in the Ember inspector. I do not think this specific code is going to be used when `@glimmer/component` is used in an Ember app, but I wanted to point out this difference just in case.

PR #220, opened 2019-11-06
[UPDATE] Updates Glimmer-VM and Typescript

pzuraq:
Updates the Glimmer-VM, and Typescript/ember-cli-typescript. The update
to TS was necessary due to updates from the Glimmer-VM. Updating ec-typescript
to make sure that published types near the same version that it supports also
seems to make sense.


mike-north:
Following a discussion w/ @pzuraq , here are some of my suggestions to help (partially/marginally) insulate consumers of this library from type-related breakage (mostly due to the large number of [breaking changes](https://github.com/microsoft/TypeScript/wiki/Breaking-Changes) associated with the 2.9.x -> 3.5.x TS compiler bump)

- Add some positive and negative test cases for the types, using either https://github.com/SamVerschueren/tsd or https://github.com/microsoft/dtslint
    - dtslint will require some special configuration, because by default it always runs tests against `typescript@next` which can be noticeably unstable at times
    - tsd does not have any built-in functionality for testing a range of compiler versions like [`dtslint does`](https://github.com/microsoft/dtslint/#specify-a-typescript-version)  
- Use [api-extractor](https://www.npmjs.com/package/@microsoft/api-extractor) to detect unintentional exports and changes to the public API surface (i.e., accidental leakage of private types through the public API)
- Create an ember-try-like script that runs the aforementioned test suite with each version of the TS compiler that this library aims to support (keeping in mind that developers who are not writing TS, but author code in a ts-aware editor are often impacted by breakage around type information)
    - I would suggest at least supporting TS 3.3.x because
         - It's the last release before some major performance regressions were introduced in TS 3.4.x.
         - TS >= 3.5 [included some breaking changes that were more difficult to adapt to than usual](https://github.com/microsoft/TypeScript/issues/33272)

pzuraq:
To be clear, prior to this change, types were being compiled with TS 3.2.3, not 2.9.x. This is part of the reason I bumped the version of ec-typescript, the actual compiled types for the built packages are currently supplied by `broccoli-typescript-compiler` (same across the whole monorepo), and we're just using `ec-typescript` to strip types with Babel.

Overall though, I agree with the rest of the points, I think it's good idea to add type tests now since they're effectively going to be part of the public API for these packages. Will add them to this PR, _before_ the upgrade commits, so we can ensure no breakage is happening.

mike-north:
> Overall though, I agree with the rest of the points, I think it's good idea to add type tests now since they're effectively going to be part of the public API for these packages. Will add them to this PR, before the upgrade commits, so we can ensure no breakage is happening.

test/types/component-test.ts
```diff
@@ -0,0 +1,76 @@
...
+  },
+  debugName: unknown,
+  didInsertElement: unknown,
+  didUpdate: unknown,
+  element: unknown,
+  isDestroying: unknown,
+  isDestroyed: unknown,
+  willDestroy: unknown,
+
+  // These are not public API, but technically part of the shape
```
tomdale: What is the rationale for including type tests for these?

pzuraq: There's no way to _not_ include them in the type check, unfortunately, it would throw an error since they exist on the object.

PR #235, opened 2020-02-08
[BREAKING] Glimmer.js 2.0

pzuraq:
This PR refactors the codebase and public APIs of Glimmer.js significantly in preparation for Glimmer.js v2.0. It includes the following changes:

### Removals

- `@glimmer/app-compiler`
- `@glimmer/application-test-helpers`
- `@glimmer/application`
- `@glimmer/compiler-delegates`
- `@glimmer/test-utils`

### Additions

- New package: `@glimmer/core`:
  - `renderComponent` - the new primary public API for starting a Glimmer application. You pass a component definition to it as the first parameter, and an element or options hash as the second, and it renders this component.

    ```js
    renderComponent(MyComponent, { element: document.querySelector('#app') });
    ```

  - `setComponentTemplate` and `createTemplate` - New base API for defining a component's template, including imports:
  
    ```js
    setComponentTemplate(
      MyComponent,
      createTemplate(
        { MyOtherComponent },
        '<MyOtherComponent/>'
      )
    );
    ```

  - A component manager that matches Ember's custom component manager API

  - A basic modifier manager, which still needs to be updated to match Ember's (in a followup PR)

- New package: `@glimmer/modifier`:
  - An importable `{{on}}` modifier that matches Ember
  - An `@action` decorator which matches the binding semantics of Ember's (but not the other behaviors)

- New package: `@glimmer/helper`:
  - A basic API for writing simple functional helpers. This will likely be updated before the final v2.0 release.

- New package: `@glimmer/babel-plugin-glimmer-env`:
  - Provides a super simple wrapper around `babel-plugin-debug-macros` to minimize setup cost, and provide debug stripping for `@glimmer/env`. This is critical now that glimmer-vm ships with the `DEBUG` flag intact.

- New package: `@glimmer/babel-plugin-strict-template-precompile`
  - Provides the compile-time transform for `createTemplate()`

### Updates

- `@glimmer/component`
  - Removed `didInsertElement`, `bounds`, and other unnecessary APIs from GlimmerComponent
  - Minor tweaks to types to make them more accurate: `GlimmerComponent<T = object>` -> `GlimmerComponent<Args extends {} = {}>`

- `@glimmer/blueprint`
  - Updated the output of the blueprint to remove ember-cli and use webpack instead. Provides a much more minimal setup, that is less opinionated. Still supports TS out of the box (but defaults to JS), and contains some basic configuration.

- Overall
  - Removes the broccoli build chain in favor of a webpack build for local development and tests, and pure TS build for publishing.
  - Removes TSLint, adds ESLint with Typescript presets and configuration
  - Adds Lerna for publishing management
  - Adds an `example-apps` directory, with the future intention being to add more complex examples including the Glimmer playground.
  
### Next Steps

- Update the Glimmer VM to the latest
- Sync ModifierManager with Ember's impl
- Add ExperimentalHostMeta
- Add ExperimentalHelperManager
- Add moar tests


packages/@glimmer/core/src/references.ts
```diff
@@ -229,26 +230,12 @@ export class UpdatableReference<T> extends ComponentPathReference<T> {
...
   }
 }
-
-export class TemplateOnlyComponentDebugReference extends ConstReference<void> {
-  constructor(protected name: string) {
-    super(undefined);
-  }
-
-  get(propertyKey: string): PathReference<unknown> {
-    throw new Error(
```
chadhietala: Is this built-in now?

packages/@glimmer/core/src/render-component/index.ts
```diff
@@ -0,0 +1,130 @@
...
+): TemplateIterator {
+  const runtime = CustomJitRuntime(resolver, context, env);
+  const builder = clientBuilder(runtime.env, {
+    element,
+    nextSibling: null,
+  } as GlimmerCursor);
+
+  const handle = resolver.registerRoot(ComponentClass);
+
+  return renderJitComponent(runtime, builder, context, 0, handle, dictToReference(componentArgs));
```
chadhietala: How would this evolve if we want to support AoT mode with `renderComponent`? One thing that <= 1.0.0 did was forced the application to decide how the VM worked e.g. AOT vs JIT, Sync Render vs Async Renderer. It seems like we would have to cut 3.0.0 to add that feature.

packages/@glimmer/core/test/interactive/fn-test.ts
```diff
@@ -0,0 +1,149 @@
+// import Component from '@glimmer/component';
```
chadhietala: Delete?

packages/@glimmer/core/test/non-interactive/render-test.ts
```diff
@@ -0,0 +1,291 @@
...
+      createTemplate(
+        { HelloWorld },
+        '<HelloWorld @name={{this.salutation}} as |name|>{{name}}</HelloWorld>'
+      )
+    );
+
+    assert.equal(await render(MainComponent), 'Glimmer!');
+  });
+
+  // [true, false].forEach((pred) => {
```
chadhietala: Delete?

packages/@glimmer/core/src/references.ts
pzuraq: Yes, but there have been more changes after the last VM upgrade in Ember. The plan is to followup with another PR to upgrade the VM again here, and rewrite some more things and move them upstream.

packages/@glimmer/blueprint/files/package.json
```diff
@@ -7,33 +7,112 @@
...
+      }
+    ]
+  },
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
+  "browserslist": {
+    "production": [
+      "last 2 Edge versions",
```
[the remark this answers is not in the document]
pzuraq: That’s a good question, we should check. My belief was that since Edge is evergreen, the “last 2” would continue to roll forward. I don’t think we should support the legacy edge once enough Chromium versions have been released.

packages/@glimmer/core/src/render-component/index.ts
pzuraq: We should discuss this more for sure before we cut the final v2 release. I know @wycats and I were talking about trying to unify the AOT and JIT models more, and we’ll have to do that to support lazy bundling and such. 

I wonder if it could be more of a compiler decision, and if we could expand the `renderComponent` API to support additional metadata for AOT compilation without needing to make breaking changes.

packages/@glimmer/core/test/interactive/fn-test.ts
pzuraq: Ah, actually the intention here is to add these tests back in followups. I commented them out for the time being so tests would pass, but these are still useful tests.

packages/@glimmer/core/test/non-interactive/render-test.ts
pzuraq: Same as above, these will be valid tests added back in followups.

lifeart:
One silly question, it it possible to return app instance in `renderComponent` promise resolution?

ref: https://github.com/tomdale/glimmer-lite-donut/issues/45

pzuraq:
@lifeart I've been thinking about that a bit, I think it would actually be good to return a public API of sorts. Specifically, I'd like the ability to `destroy` the instance so that manually managed instances can be cleaned up properly, but that could be expanded in the future.

I think we can do that as a separate PR though, either before v2 is released or sometime after. What use cases were you thinking of that you'd like to support?

lifeart:
@pzuraq use cases:

* web components api integration (properties, attributes, destroying) support.

https://gist.github.com/lifeart/2c4666844a11b1e8536a32d6c776aacc

`attributeChangedCallback`, `connectedCallback`, `disconnectedCallback`, `adoptedCallback`

* multiple glimmer applications in one page with single state management (including it's lifecyle) just like multiple Vue apps in one page, to have ability integrate glimmer with legacy systems step-by step.

If we have external data source or business logic, it will be great to have access to root component instance to invoke some methods or set properties to it without hacking.

packages/@glimmer/core/src/render-component/index.ts
chadhietala: Yea I would assume that having first class values makes a lot of the resolver indirection.

I think the thing I think we should do here is to require `renderComponent` to take the mode specific bits. Its not a big deal for the server-side have defaults but I think we should continue to have code for configuration to ensure that rendering can happen different ways, but not pay the tax of downloading the code for the unused modes.

For example we could do something like capabilities:

```js
renderComponent(Foo, {
  element: document.getElementById('foo'),
  mode: JIT()
});
```

Then we would have mode mappings like the following:

```js
const CURRENT_MODE = new WeakMap();

const JITSetup = {
    renderer: renderJitComponent,
    runtime: CustomJitRuntime,
     // ...
}

const AoTSetup = {
    renderer: renderAoTComponent,
    runtime:   AotRuntime,
     // ...
}

const JITMode =  {
  mode: 'JIT',
};

// All the JIT bits
function JIT() {
  CURRENT_MODE.set(JITMode, JITSetup);
  return JITMode;
}

const AoTMode =  {
  mode: 'AoT',
};

// All the JIT bits
function AoT() {
  CURRENT_MODE.set(AoTMode, AoTSetup);
  return AoTMode;
}

renderComponent(Component, options) {
  const { mode } = options.mode;
  const { Renderer, Runtime } = CURRENT_MODE.get(options.mode);
  
  if (mode === 'JIT) {
    // JIT specific setup
  } else {
    // AoT specific setup 
  }
}

```

packages/@glimmer/core/test/interactive/fn-test.ts
chadhietala: Maybe use `todo` then?

chadhietala:
@pzuraq looks like the publishing script was removed. Any intention of bringing it back in this PR?

pzuraq:
I believe we can just use lerna publish with this setup, as it builds the packages directly in the same directory rather than a new `dist` folder. I found this pretty convenient for local linking as well.

packages/@glimmer/ssr/test/render-options-tests.ts
```diff
@@ -0,0 +1,26 @@
+import Component from '@glimmer/component';
+import { setComponentTemplate, createTemplate } from '@glimmer/core';
+import HTMLSerializer from '@simple-dom/serializer';
+import voidMap from '@simple-dom/void-map';
+import { SerializableNode } from '@simple-dom/interface';
+import { renderToString, RenderOptions } from '..';
+
+QUnit.module('@glimmer/ssr rendering', () => {
```
chadhietala: I don't think these tests are running at all.

pzuraq: It does if you run `testem ci`, which launches the Node process to run those on the command line. This setup was copied from `glimmer-lite-donut`, it could probably be a bit more explicit with a `yarn test:node` command, though I'm not sure how easy or hard it would be to get Testem to work that way.

It's also debatable if the SSR code should run in the browser, I could add the tests to the browser tests as well. I think it's probably useful, `renderToString` isn't really Node specific.

packages/@glimmer/core/test/interactive/fn-test.ts
rwjblue: @chadhietala `QUnit.todo` does not work with testem

chadhietala: 💩

packages/@glimmer/core/src/template.ts
```diff
@@ -0,0 +1,37 @@
+import { SerializedTemplateWithLazyBlock, Dict } from '@glimmer/interfaces';
+import { TemplateMeta } from './managers/component/custom';
+
+const TEMPLATE_MAP = new WeakMap<object, SerializedTemplateWithLazyBlock<TemplateMeta>>();
+const getPrototypeOf = Object.getPrototypeOf;
+
+// This is provided by the `babel-plugin-strict-template-precompile` plugin
+export declare function createTemplate(
```
chadhietala: From our chat, can we just make this a function that throws and tells people to use the plugin.

index.html
```diff
@@ -0,0 +1,10 @@
+<!doctype html>
```
rwjblue: Why is this file here? Shouldn't be in a subdir?

package.json
```diff
@@ -3,65 +3,55 @@
...
+    "build": "bin/build.js",
+    "start": "webpack-dev-server",
+    "clean": "bin/clean.js",
     "changelog": "lerna-changelog",
-    "start": "ember serve",
-    "test": "ember test && ember test --env production",
-    "test:node": "bin/run-node-tests.js",
-    "test:ember": "yarn workspace @glimmer/component ember try:one",
+    "test": "yarn lint && testem ci && yarn test:babel-plugins",
+    "test:babel-plugins": "mocha -r esm --timeout 5000 test/babel-plugins.js",
```
rwjblue: Should depend on `esm` no?

PR #238, opened 2020-02-14
[REFACTOR] Updates to Glimmer-VM 0.47

pzuraq:


packages/@glimmer/blueprint/files/webpack.config.js
```diff
@@ -42,7 +42,6 @@ module.exports = () => {
       rules: [
         {
           test: /\.(js|mjs|ts)$/,
-          include: [path.resolve(__dirname, 'src'), path.resolve(__dirname, 'tests')],
```
pzuraq: It turns out with this included, `node_modules` will not get transpiled at all.

PR #239, opened 2020-02-14
[REFACTOR] Removes most of the remaining custom reference code.

pzuraq:
Removes most of the remaining custom reference code, with the exception
of helper references (which will be factored out soon).

Also removes the tracking/rendering assertion for the time being,
pending discussion on whether or not the assertion continues to make
sense. If the discussion concludes it does, we can add it back before
the v2 release, with more thorough integration testing that actually
tests the expected behavior, rather than unit tests.


chadhietala:
I think we added that assertion because it wasn't really obvious when working on an app that a field wasn't being tracked.

pzuraq:
The issue is that with autotracking, we can no longer provide guarantees about this assertion. Basically, as soon as you add a getter, we can’t wrap any properties that getter accesses.

Personally, I think this is worse DX. A dev might encounter the error the first time with a normal property, think “Oh, Glimmer will tell me if I need to add the decorator, neat!”, and then be confused and upset when that doesn’t work in all cases.

PR #240, opened 2020-02-14
[FEATURE] Adds Owner

pzuraq:
Adds a `owner` argument that can be passed to `renderComponent`. This
value is provided to every manager in that render tree via the manager
factory:

```js
setComponentManager((owner) => new CustomManager(owner), {});
```

Conceptually, the owner here is similar to the owner in Ember.js.
Glimmer.js creates a new manager for each unique owner that it receives,
and managers can use the `getOwner` and `setOwner` methods to assign it
to passed objects. The owner is an opaque value, so it's up to the
embedding environment to decide what its API is. If no owner is passed,
a default owner is used, and throws an error if the user attempts to use
it for anything.

---

Breaking changes:

* Removes support for Ember prior to 3.8, which used a (now deprecated) version of component manager association that relied on Ember's resolver.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,4 +44,5 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-release
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.4
+    - env: EMBER_TRY_SCENARIO=ember-lts-3.12
+    - env: EMBER_TRY_SCENARIO=ember-lts-3.16
--- a/packages/@glimmer/component/addon/index.ts
+++ b/packages/@glimmer/component/addon/index.ts
@@ -1,21 +1,12 @@
 import { DEBUG } from '@glimmer/env';
 import ApplicationInstance from '@ember/application/instance';
 import { setComponentManager } from '@ember/component';
-import { gte } from 'ember-compatibility-helpers';
 
 import GlimmerComponentManager from './-private/ember-component-manager';
-import GlimmerComponent from './-private/component';
+import _GlimmerComponent from './-private/component';
 import { setOwner } from '@ember/application';
 
-if (gte('3.8.0-beta.1')) {
-  setComponentManager((owner: ApplicationInstance) => {
-    return new GlimmerComponentManager(owner);
-  }, GlimmerComponent);
-} else {
-  setComponentManager('glimmer', GlimmerComponent);
-}
-
-export default class extends GlimmerComponent {
+export default class GlimmerComponent extends _GlimmerComponent {
   constructor(owner, args) {
     super(owner, args);
 
@@ -28,3 +19,7 @@ export default class extends GlimmerComponent {
     setOwner(this, owner);
   }
 }
+
+setComponentManager((owner: ApplicationInstance) => {
+  return new GlimmerComponentManager(owner);
+}, GlimmerComponent);
--- a/packages/@glimmer/core/src/managers/component/custom.ts
+++ b/packages/@glimmer/core/src/managers/component/custom.ts
@@ -13,11 +13,15 @@ import {
   TemplateOk,
   Environment,
   CompilableProgram,
+  DynamicScope,
 } from '@glimmer/interfaces';
 import { PathReference, ComponentRootReference } from '@glimmer/reference';
 import { Tag, isConst, createTag, consume } from '@glimmer/validator';
+import { OWNER_KEY, DEFAULT_OWNER } from '../../owner';
 
 import { unwrapTemplate } from '@glimmer/opcode-compiler';
+import { getComponentManager } from '..';
+import { TemplateMeta } from '../../template';
 
 export const VM_CAPABILITIES: VMComponentCapabilities = {
   createInstance: true,
@@ -53,7 +57,7 @@ export function capabilities(
     'Invalid component manager compatibility specified'
   );
 
-  const updateHook = managerAPI === '3.13' ? Boolean(options.updateHook) : true;
+  const updateHook = managerAPI !== '3.4' ? Boolean(options.updateHook) : true;
 
   return {
     asyncLifecycleCallbacks: Boolean(options.asyncLifecycleCallbacks),
@@ -130,10 +134,6 @@ export interface ComponentArguments {
   named: Dict<unknown>;
 }
 
-export interface TemplateMeta {
-  scope: () => Dict<unknown>;
-}
-
 export interface Destroyable {
   destroy(): void;
 }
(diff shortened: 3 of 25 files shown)


rwjblue:
I don't fully grok why we need a new concept here, vs using the pre-existing `getOwner` system?

packages/@glimmer/component/addon/index.ts
```diff
@@ -1,21 +1,12 @@
...
-import GlimmerComponent from './-private/component';
+import _GlimmerComponent from './-private/component';
 import { setOwner } from '@ember/application';
 
-if (gte('3.8.0-beta.1')) {
-  setComponentManager((owner: ApplicationInstance) => {
-    return new GlimmerComponentManager(owner);
-  }, GlimmerComponent);
-} else {
-  setComponentManager('glimmer', GlimmerComponent);
```
rwjblue: Can you make sure this PR is labeled as breaking, and add a small section (under the existing PR description) that explains exactly _what_ is breaking (I think this might be the only thing?).

pzuraq:
This also removes support for Ember < 3.8 as well, removing the legacy component managers for Glimmer components

packages/@glimmer/core/src/managers/component/custom.ts
```diff
@@ -42,18 +46,18 @@ export interface Capabilities {
 
 export type OptionalCapabilities = Partial<Capabilities>;
 
-export type ManagerAPIVersion = '3.4' | '3.13';
+export type ManagerAPIVersion = '3.4' | '3.13' | 'experimental';
```
rwjblue: I think we can probably remove the `experimental` one here

packages/@glimmer/core/src/owner.ts
```diff
@@ -0,0 +1,27 @@
+import { DEBUG } from '@glimmer/env';
+
+const OWNER_MAP = new WeakMap<{}, unknown>();
+
+export const OWNER_KEY = `__OWNER_${Math.floor(Math.random() * Date.now())}__`;
```
rwjblue: Can we use a "real" symbol for this? Glimmer.js doesn't support IE11 (glimmer-vm does), so I think we can use `Symbol` natively here...

packages/@glimmer/core/src/owner.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export let DEFAULT_OWNER = {};
```
rwjblue: I'm curious here, do you think this `DEFAULT_OWNER` solution is overall better than just passing `null`? I think the errors will be really nice with the current implementation, just wondering if you had considered "just pass null"...

packages/@glimmer/core/src/owner.ts
```diff
@@ -0,0 +1,27 @@
...
+    get(): never {
+      throw new Error(OWNER_ERROR);
+    },
+    set(): never {
+      throw new Error(OWNER_ERROR);
+    }
+  });
+}
+
+export function getOwner<T = unknown>(obj: object): T {
```
rwjblue: I don't think it will matter too much, but `obj` just needs to be a valid weakmap key and that _would_ allow functions.

packages/@glimmer/core/test/non-interactive/render-test.ts
```diff
@@ -165,28 +166,26 @@ module(`[@glimmer/core] non-interactive rendering tests`, () => {
   //   assert.equal(await render(MainComponent), 'Hello Glimmer!');
   // });
 
-  // test('a component can inject services', async assert => {
-  //   class LocaleService {
-  //     get currentLocale(): string {
-  //       return 'en_US';
-  //     }
-  //   }
+  test('components recieve owner', async assert => {
```
rwjblue:
```suggestion
  test('components receive owner', async assert => {
```

packages/@glimmer/core/test/non-interactive/render-test.ts
```diff
@@ -165,28 +166,26 @@ module(`[@glimmer/core] non-interactive rendering tests`, () => {
...
 
-  //   const html = await render(MyComponent, {
-  //     meta: {
-  //       locale: new LocaleService(),
-  //     },
-  //   });
-  //   assert.strictEqual(html, '<h1>en_US</h1>');
-  // });
+    const html = await render(MyComponent, {
+      owner: new LocaleService(),
```
rwjblue: I think it would be generally better to keep this test somewhat representative of what we expect folks to do.

I'm thinking:

```suggestion
      owner: { 
        services: {
         locale: new LocaleService(),
        },
      },
```

packages/example-apps/basic/src/MyComponent.ts
```diff
@@ -36,8 +38,7 @@ class MyComponent extends Component {
   @tracked count = 55;
 
   get currentLocale(): string {
-    return 'zh_CN';
-    // return (getHostMeta(this) as { locale: LocaleService }).locale.currentLocale;
+    return getOwner<{ services: { locale: LocaleService } }>(this).services.locale.currentLocale;
```
rwjblue: For the example app, let's make an interface for the owner and then use it here.

packages/@glimmer/core/src/owner.ts
pzuraq: We can't use a real one unfortunately, this is for DynamicScope which only currently works with strings (at least, according to types)

.travis.yml
```diff
@@ -44,4 +44,4 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-release
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.4
```
rwjblue: Can you also add a CI run for Ember 3.16? I know it happens to be the current version, but #soon it will not be 😸

packages/@glimmer/core/src/owner.ts
pzuraq: The main reason I did it this way was so that manager instances could go into the `WeakMap` mapped to owners. If we allowed `null`, we would have to have a branch for that I suppose

packages/@glimmer/core/src/owner.ts
pzuraq: I believe `object` works this way as a type: https://www.typescriptlang.org/play/index.html?ssl=3&ssc=17&pln=1&pc=1#code/DYUwLgBA9gRgVgLmvEBjMBuAsAKF7OCAXggAoBKYgPggG8IBfDIA

rwjblue: WAT, 😞 sorry I didn't realize that

packages/@glimmer/core/src/owner.ts
rwjblue: 😭 ok

.travis.yml
rwjblue: Thanks for adding the `config/ember-try.js` entry for 3.16, I think you still need to add it here (or it will never run).

PR #243, opened 2020-02-29
[FEAT] Adds Helpers and HelperManager

pzuraq:
Changes:

* Removes the `helper` function and reference implementation from
  `@glimmer/helper`

* Adds a simple-helper manager that wraps plain functions when used as
  helpers. Plain functions are only provided positional parameters, and
  using named params with them will throw an error.

* Adds a custom-helper manager that is exposed publicly via the
  `setHelperManager` API. Currently it consists of:
  - Hooks
    - `createHelper`
    - `updateHelper`
    - `destroyHelper`
    - `getState`
  - `capabilities`
    - `disableAutoTracking`
    - `destructor`

* Adds a functional implementation of the `{{fn}}` helper to
  `@glimmer/helper`.

* Adds a built-in inline `{{if}}` helper, since that is part of the
  syntax moreso than an actual helper.

* Adds some tests for `toBool` behavior to ensure we match Ember's
  behavior (for interop).


package.json
```diff
@@ -66,5 +79,8 @@
       "internal": ":house: Internal",
       "deprecation": ":warning: Deprecation"
     }
+  },
+  "dependencies": {
+    "amd-name-resolver": "https://github.com/ember-cli/amd-name-resolver.git"
```
rwjblue: This seems odd, why do we need to use this from git?

rwjblue:
Needs a rebase (no that the owner stuff is landed), probably drop the realms commit?

packages/@glimmer/core/src/render-component/resolvers.ts
```diff
@@ -76,12 +81,11 @@ export class CompileTimeResolver implements ResolverDelegate {
 
   lookupHelper(name: string, referrer: TemplateMeta): Option<number> {
     const scope = referrer.scope();
-    const Helper = scope[name] as GlimmerHelper;
+    const Helper = scope[name] as HelperDefinition || builtInHelpers[name];
```
rwjblue: Built-in's should always win, this order seems wrong

packages/@glimmer/helper/src/fn.ts
```diff
@@ -0,0 +1,11 @@
+import { DEBUG } from '@glimmer/env';
+import { assert } from '@glimmer/util';
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export default function fn(fn: (...args: unknown[]) => unknown, ...args: unknown[]) {
+  if (DEBUG) {
+    assert(typeof fn === 'function', 'fn must receive a function as its first parameter');
+    assert(args.length > 0, 'fn must receive at least one argument to pass to the function, otherwise there is no need to use fn.');
+  }
+  return fn.bind(null, ...args);
```
rwjblue: We should use the "untouchable this" thing

https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/lib/utils/untouchable-this.ts

https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/lib/helpers/fn.ts#L8

packages/@glimmer/core/src/environment/to-bool.ts
```diff
@@ -0,0 +1,7 @@
+export default function toBool(predicate: unknown): boolean {
+  if (Array.isArray(predicate)) {
+    return (predicate as { length: number }).length !== 0;
```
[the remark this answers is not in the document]
pzuraq: This might have been a holdover from Ember, since it uses `isEmberArray`

packages/@glimmer/core/src/managers/util.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      if (parsed !== null && parsed < positional.length) {
+        const ref = positional.at(parsed);
+        consume(ref.tag);
+
+        return ref.value();
+      } else if (prop === 'length') {
+        consume(positional.tag);
+        return positional.length;
+      }
```
rwjblue: Future optimization: only call `convertToInt` when `prop` is not `length`

packages/@glimmer/core/src/managers/util.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return (target as any)[prop];
+    },
+
+    has(_target, prop) {
+      return typeof prop === 'number' && prop < positional.length;
```
rwjblue: This will fail for `'0' in this.args.positional`, do we care?

packages/@glimmer/core/src/managers/util.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const namedTarget = {};
+  const positionalTarget: unknown[] = [];
+
+  if (DEBUG) {
+    const setHandler = function(_target: unknown, prop: symbol | string | number): never {
+      throw new Error(
+        `You attempted to set ${String(
+          prop
+        )} on the arguments of a component, helper, or modifier. Arguments are immutable and cannot be updated directly, they always represent the values that is passed down. If you want to set default values, you should use a getter and local tracked state instead.`
```
rwjblue: When you call `argsProxyFor`, can you pass another argument for the `type`? Ideally we would have a way to figure out the name of the thing too, but that seems good for a future iteration.

packages/@glimmer/core/src/managers/util.ts
```diff
@@ -0,0 +1,102 @@
...
+      );
+    };
+
+    const forInDebugHandler = (): never => {
+      throw new Error(
+        'ownKeys was called on the positional arguments array for a component, helper, or modifier. This function is a low-level function that should not need to be called for positional argument arrays. You may be attempting to iterate over the array using for...in instead of for...of, or directly trying to lookup the keys of the array using Object.keys, which is not supported'
```
rwjblue: We should probably:

* not say "ownKeys" (AFAICT this isn't a thing people would every type other than for a proxy trap), maybe `Object.keys` instead?
* say exactly what type of object it is (`argsProxyFor('helper', capturedArgs)`)

packages/@glimmer/core/test/interactive/helper-test.ts
```diff
@@ -43,12 +183,244 @@ module('Modifier Tests', () => {
...
+          positional: [string];
+          named: {
+            greeting: string;
+          };
+        };
+
+        setup(): void {
+          // entangle a value that changes
+          this.args.positional[0];
+          setupCount++;
```
rwjblue: Instead of incrementing, can you push into an array (that we assert with deepEqual)?

packages/@glimmer/core/test/non-interactive/render-test.ts
```diff
@@ -127,29 +111,44 @@ module(`[@glimmer/core] non-interactive rendering tests`, () => {
...
-  //       pred = pred;
-  //       salutation = 'Glimmer';
-  //       alternative = 'Glimmer.js';
-  //     }
+  [true, false].forEach(pred => {
+    test(`can use inline if - ${pred}`, async function(assert) {
+      class Main extends Component {
+        pred = pred;
+        salutation = 'Glimmer';
+        alternative = 'Glimmer.js';
```
rwjblue: Can you make these getters, then when retrieved push into an array that we assert at the end? Specifically, I'd like the test to show that the flasey side is not called when the value is true.

packages/@glimmer/core/src/managers/component/custom.ts
```diff
@@ -184,79 +180,22 @@ export default class CustomComponentManager<ComponentInstance>
...
-    return new VMCustomComponentState(env, delegate, component, capturedArgs, namedArgsProxy);
+    return new VMCustomComponentState(env, delegate, component, capturedArgs, argsProxy);
   }
 
   update({
     delegate,
     component,
-    args,
-    namedArgsProxy,
+    argsProxy,
```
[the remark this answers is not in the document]
pzuraq: Definitely a point to consider in the future, but for now I think the extra clarity is better. We can always optimize later on.

packages/@glimmer/core/src/managers/helper.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  if (disableAutoTracking === false) {
+    getValue = trackedMemoize(getValue);
+    createOrUpdate = trackedMemoize(createOrUpdate);
+  } else {
+    helperFn = (): unknown => {
+      let ret;
+      untrack(() => (ret = helperFn()));
+      return ret;
+    };
```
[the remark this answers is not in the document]
pzuraq: Decided to remove `disableAutoTracking` as a feature, we don't actually need it as a capability at the moment and can add it in the future if needed.

PR #245, opened 2020-03-27
[FEAT] Modifier Managers

pzuraq:
Adds modifier managers that match Ember's modifier manager API. They can
be defined by importing `setModifierManager` from `@glimmer/core`:

```ts
import {
  modifierCapabilities,
  setModifierManager,
} from '@glimmer/core';
```

Also allows users to use simple functions as modifiers:

```js
function scroll(element, position) {
  element.scrollY = position;

  return () => element.scrollY = 0;
}
```

The function receives the element as the first argument, and any
positional arguments after that. It can return a destructor, which will
be called whenever the modifier is destroyed, or before it is updated
and the modifier is run again.

Also adds a `TrackedObject` implementation for tests, which makes life
for testing a bit easier.

### DIFF
--- a/packages/@glimmer/core/index.ts
+++ b/packages/@glimmer/core/index.ts
@@ -16,6 +16,13 @@ export {
 export { Args as CapturedArgs } from './src/interfaces';
 
 export {
+  ModifierManager,
+  ModifierDefinition,
+  capabilities as modifierCapabilities,
+  Capabilities as ModifierCapabilities,
+} from './src/managers/modifier';
+
+export {
   HelperManager,
   HelperDefinition,
   capabilities as helperCapabilities,
--- a/packages/@glimmer/core/src/managers/modifier.ts
+++ b/packages/@glimmer/core/src/managers/modifier.ts
@@ -1,6 +1,20 @@
-import { assert } from '@glimmer/util';
+import { DEBUG } from '@glimmer/env';
+import {
+  ModifierManager as VMModifierManager,
+  VMArguments,
+  CapturedArguments,
+  Destroyable,
+  DynamicScope,
+} from '@glimmer/interfaces';
+import { Tag, createUpdatableTag, track, untrack, combine, update } from '@glimmer/validator';
+import { assert, debugToString } from '@glimmer/util';
 import { SimpleElement } from '@simple-dom/interface';
 import { Args } from '../interfaces';
+import debugRenderMessage from '../utils/debug';
+import { argsProxyFor } from './util';
+import { getModifierManager } from '.';
+import { OWNER_KEY } from '../owner';
+import { VMModifierDefinitionWithHandle } from '../render-component/vm-definitions';
 
 ///////////
 
@@ -10,7 +24,7 @@ export interface Capabilities {
 
 export type OptionalCapabilities = Partial<Capabilities>;
 
-export type ManagerAPIVersion = '3.4' | '3.13';
+export type ManagerAPIVersion = '3.13';
 
 export function capabilities(
   managerAPI: ManagerAPIVersion,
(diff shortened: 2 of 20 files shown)


packages/@glimmer/core/index.ts
```diff
@@ -15,6 +15,13 @@ export {
 
 export { Args as CapturedArgs } from './src/interfaces';
 
+export {
+  ModifierManager,
+  ModifierDefinition,
+  capabilities as modifierCapabilities,
+  Capabilities as ModifierCapabilities,
+} from './src/managers/modifier';
```
rwjblue: Thank you for not using `export *` 👯‍♂

packages/@glimmer/core/src/managers/modifier.ts
```diff
@@ -25,12 +39,152 @@ export function capabilities(
...
+    definition: ModifierDefinition<ModifierStateBucket>,
+    args: VMArguments,
+    dynamicScope: DynamicScope
+  ): CustomModifierState<ModifierStateBucket> {
+    const owner = dynamicScope.get(OWNER_KEY).value() as object;
+    let delegate = getModifierManager(owner, definition);
+
+    if (delegate === undefined) {
+      if (DEBUG) {
+        assert(
```
rwjblue: Curious, do these get stripped properly on their own or do they have to be within an `if (DEBUG) {`? Also, if they do have to be in an `if (DEBUG) {` what happens to the import?

packages/@glimmer/core/test/interactive/modifier-test.ts
```diff
@@ -24,23 +70,276 @@ module('Modifier Tests', () => {
...
     );
 
-    const button = element.querySelector('button')!;
-    button.click();
+    click('button');
 
-    await settled();
     assert.strictEqual(
-      element.innerHTML,
+      await settled(),
```
rwjblue: Seems somewhat surprising that `settled` returns the HTML

packages/@glimmer/core/test/utils/dom.ts
```diff
@@ -0,0 +1,11 @@
+export function find(selector: string): Element {
+  const element = document.getElementById('qunit-fixture')!;
+  return element.querySelector(selector)!;
+}
+
+export function click(selector: string): void {
+  const element = document.getElementById('qunit-fixture')!;
+  const clickable = element.querySelector(selector)! as HTMLButtonElement;
+
+  clickable.click();
```
rwjblue: Should be using dispatch event here probably. Seems fine for now, it just reminds me that we really need to genericize the `@ember/test-helpers` DOM utilities to work without Ember / Glimmer.

packages/@glimmer/core/test/utils/tracked-object.js
```diff
@@ -0,0 +1,58 @@
+import { consume, tagFor, dirtyTagFor } from '@glimmer/validator';
```
rwjblue: Why not just author this in .ts?

packages/@glimmer/tracking/src/tracked.ts
```diff
@@ -139,15 +139,6 @@ function descriptorForField<T extends object, K extends keyof T>(
...
   };
 }
-
-// eslint-disable-next-line @typescript-eslint/no-empty-function
-let propertyDidChange = function(): void {};
-
-// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export function setPropertyDidChange(cb: () => void) {
-  propertyDidChange = cb;
-}
```
rwjblue: Can you help me understand this change? Do we not need to be able to customize this anymore?

packages/@glimmer/core/test/interactive/modifier-test.ts
pzuraq: This is the pattern established by the `render()` test helper. I agree, the more I work with it the more I think we need something a bit better. Ideally we would get QUnit.dom working in this repo, and allow `render()` to return the result (along with a way to destroy the result).

packages/@glimmer/core/src/managers/modifier.ts
pzuraq: I believe the import should be tree-shaken, which is why I figured this pattern would be alright for now. We currently only replace `DEBUG`.

packages/@glimmer/tracking/src/tracked.ts
pzuraq: This was a mistake I made earlier, `propertyDidChange` already exists in the `getter`/`setter` setup from `@glimmer/validator`. This meant there were two different `propertyDidChange` callbacks being fired.

packages/@glimmer/core/test/utils/tracked-object.js
pzuraq: It is impossible, you can't have `new TrackedObject` return anything but a `TrackedObject` if you author it as a TS class, and there's no way to tell TS that this `TrackedObject` should get the same keys as the object passed to it. @chriskrycho and I noodled on this for a while with tracked-built-ins, and this was the only way I found you could get the expected behavior.

That aside, Proxy itself is very difficult to type correctly. I haven't been able to figure out a way to type the handlers of a Proxy without casting to `any` multiple times yet. So, I'm not sure TS would help us much with the implementation here.

pzuraq:
I made an issue to track some of the ongoing problems/pain points with the current testing harness: https://github.com/glimmerjs/glimmer.js/issues/246

PR #247, opened 2020-03-27
chore: extend example app to js-benchmark

lifeart:


packages/example-apps/basic/src/MyTable.ts
```diff
@@ -0,0 +1,133 @@
...
+
+import { fn } from './utils/fn-helper';
+
+import TableRow from './TableRow';
+import BsButton from './BsButton';
+
+
+export default class MyTable extends Component {
+  @tracked
+  id = 1;
```
rwjblue:
```suggestion
  @tracked id = 1;
```

packages/example-apps/basic/src/MyTable.ts
```diff
@@ -0,0 +1,133 @@
...
+  @tracked
+  data = [];
```
rwjblue:
```suggestion
  @tracked data = [];
```

packages/example-apps/basic/src/MyTable.ts
```diff
@@ -0,0 +1,133 @@
...
+  @tracked
+  selected = undefined;
```
rwjblue:
```suggestion
  @tracked selected = undefined;
```

packages/example-apps/basic/src/utils/fn-helper.ts
```diff
@@ -0,0 +1,9 @@
+import { helper } from './helper-with-services';
```
rwjblue: Why do you need `./helper-with-services` here? Seems like a simple function would work 🤔

packages/example-apps/basic/src/utils/fn-helper.ts
```diff
@@ -0,0 +1,9 @@
...
+
+
+export const fn = helper(function([method, ...args]) {
+  return  (...params)  => {
+      let el = [].concat(args, params);
+      return method(...el);
+  }
+});
```
rwjblue:
```suggestion
export function fn([method, ...args]) {
  return  (...params)  => {
      let el = [].concat(args, params);
      return method(...el);
  }
};
```

packages/example-apps/basic/src/utils/fn-helper.ts
lifeart: plain function not working for me :/

lifeart: do I need to use `setHelperManager` ?

pzuraq: You should be able to import `fn` directly:

```
import { fn } from '@glimmer/helper';
```

It is implemented as a plain functoin.

packages/example-apps/basic/src/utils/fn-helper.ts
pzuraq: It would actually be:

```js
export function fn(method, ...args) {
  return  (...params)  => {
    let el = [].concat(args, params);
    return method(...el);
  }
}
```

The outer function only receives positional parameters

lifeart:
Dear developer, I'm closing this PR, looks like it does not have use cases.
For real-life glimmer usage check https://github.com/lifeart/glimmerx-workshop

PR #250, opened 2020-03-30
Add automated release setup.

rwjblue:
* The process is outlined in `RELEASE.md`.
* Removes `lerna` and `lerna.json` (should not be needed).


CHANGELOG.md
```diff
@@ -1,5 +1,3 @@
-# Changelog
```
rwjblue: This is removed because all future additions to the changelog are pushed onto the beginning.

PR #256, opened 2020-04-01
[FEAT] Adds a manual `precompileTemplate` API

pzuraq:
This API allows users to precompile templates themselves, so they can
build babel transforms on top of them. The API returns a string value
that users can simply insert into their output.

### DIFF
--- a/packages/babel-plugins/@glimmer/babel-plugin-strict-template-precompile/index.js
+++ b/packages/babel-plugins/@glimmer/babel-plugin-strict-template-precompile/index.js
@@ -1,4 +1,7 @@
 const { precompile } = require('@glimmer/compiler');
+const { parse } = require('@babel/parser');
+const { default: generate } = require('@babel/generator');
+const t = require('@babel/types');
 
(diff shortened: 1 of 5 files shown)


packages/babel-plugins/@glimmer/babel-plugin-strict-template-precompile/index.js
```diff
@@ -1,4 +1,7 @@
 const { precompile } = require('@glimmer/compiler');
+const { parse } = require('@babel/parser');
+const { default: generate } = require('@babel/generator');
```
rwjblue, line 3: p1/5. 

I _vaguely_ prefer this (I think it is easier to parse at a glance):

```suggestion
const generate = require('@babel/generator').default;
```

pzuraq: I actually tried to do that, but it wasn't working, said `generate` was not a function. I think it might be due to the way they write/transpile it for publishing.

Issue #253, opened 2020-03-31
Example in blueprint fails in v2.0.0-beta.2


josemarluedke:
The example in the blueprint has the following:

```hbs
<img src="{{this.logo}}" />
```

When running the app, we get the following error:

```
delegates.js:63 Uncaught (in promise) TypeError: Cannot set property 'href' of undefined
    at EnvironmentImpl.protocolForURL (delegates.js:63)
    at sanitizeAttributeValue (sanitized-values.js:48)
    at SafeDynamicProperty.set (dynamic.js:154)
    at NewElementBuilder.setDynamicAttribute (element-builder.js:335)
    at Object.eval [as evaluate] (dom.js:181)
    at AppendOpcodes.evaluate (opcodes.js:74)
    at LowLevelVM.evaluateSyscall (low-level.js:171)
    at LowLevelVM.evaluateInner (low-level.js:127)
    at LowLevelVM.evaluateOuter (low-level.js:119)
    at JitVM.next (append.js:398)
```
which points directly to:
https://github.com/glimmerjs/glimmer.js/blob/62a4ae41718b2138e9a35b810e16ff6879b93057/packages/%40glimmer/core/src/environment/delegates.ts#L64

josemarluedke:
I'm using Brave 1.5.115 Chromium: 80.0.3987.149.

pzuraq:
Ah, found the issue

PR #258, opened 2020-04-01
[BUGFIX] Binds EnvironmentDelegate.protocolForUrl

pzuraq:
We made a decision upstream to pluck the delegate functions and move
them around so that they could be looked up more quickly when used. In
most cases this is fine, because the delegate functions are pure and
non-stateful. The `protocolForUrl` function is not, but this was known
at the time, and the plan was to bind it like in this PR. It was an
oversight in the upgrade process.

Tests added to ensure we remember if future changes occur.

### DIFF
--- a/packages/@glimmer/core/src/environment/delegates.ts
+++ b/packages/@glimmer/core/src/environment/delegates.ts
@@ -49,16 +49,9 @@ if (DEBUG) {
 export class ClientEnvDelegate extends BaseEnvDelegate {
   isInteractive = true;
 
-  private uselessAnchor: HTMLAnchorElement;
+  private uselessAnchor = self.document.createElement('a');
 
-  constructor() {
-    super();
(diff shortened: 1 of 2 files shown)


pzuraq:
Fixes #253

packages/@glimmer/core/src/environment/delegates.ts
```diff
@@ -58,7 +58,7 @@ export class ClientEnvDelegate extends BaseEnvDelegate {
     this.uselessAnchor = self.document.createElement('a');
   }
 
-  protocolForURL(url: string): string {
+  protocolForURL = (url: string): string => {
```
rwjblue, line 61: p3/5

Couple of issues:

* In general, I'd prefer to keep the class fields together (makes scanning the file much easier)
* I think we should also make `uselessAnchor` a class field
* Once we address both of those things, we can delete the constructor 😸

PR #263, opened 2020-04-01
Update to prettier@2

rwjblue:
* Updates to prettier@ (see [blog post](https://prettier.io/blog/2020/03/21/2.0.0.html) for details).
* Fix issue with prettier linting setup (prettier was not being enforced by `eslint` prior to these changes).
* Run `yarn lint --fix` to fix all the files.

This will be easiest to review by looking at the first commit on its own, as the second commit is `eslint --fix .` basically.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,6 +14,6 @@ typescript.ts
 babel.js
 
 **/fixtures/**/*.js
-**/fixture-options/**/*.js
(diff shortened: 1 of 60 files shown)


rwjblue:
Failures are from files that should be excluded from prettier (babel plugin fixture files that are asserting the expected output from babel).

PR #265, opened 2020-04-02
Add prettier to generated project's package.json.

rwjblue:
The linting configuration leveraged `prettier`, but did not actually install it.


rwjblue:
CI status seem to be passing but not reporting to GH:

https://travis-ci.org/github/glimmerjs/glimmer.js/builds/670182031

PR #270, opened 2020-04-03
Avoid using `&&` in `package.json` scripts.

rwjblue:
* Using `&&` locks us into specific shell support requirements which we don't _really_ care about
* Using `npm-run-all` will properly use `yarn` or `npm` depending on what is in use in the project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,9 @@
     "lint": "eslint . --cache --ext .js,.ts",
     "problems": "tsc -p tsconfig.json --noEmit",
     "start": "webpack-dev-server",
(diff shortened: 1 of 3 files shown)


pzuraq:
Does this preserve ordering of the logs in general? Just don't want to get the different outputs mixed up

pzuraq:
Realized that the jobs in Actions were split up anyways, which is all I really cared about.

rwjblue:
> Does this preserve ordering of the logs in general? Just don't want to get the different outputs mixed up

Yes, it does not run them in parallel by default (though there _is_ a way to do that) so the resulting log output is still in the correct order.

PR #283, opened 2020-04-18
[Refactor] Updates tests to use tracked utility.

rahilvora:
Updates test file to use tracked utility for component properties which requires tracking.


pzuraq:
Awesome, this looks great! Thanks much 😄

PR #292, opened 2020-07-24
Check for `meta.destroy` to avoid deprecation in ember 3.20 series

jordpo:
Relates to this issue https://github.com/emberjs/ember.js/issues/19042

PR #293, opened 2020-07-27
Leverage built-in destroyables on Ember 3.20+ to avoid deprecations

rwjblue:
Avoids deprecations for `Meta.prototype.setSourceDestroyed` and `Meta.prototype.setSourceDestroying` in Ember 3.20+.

Closes #292 
Closes https://github.com/emberjs/ember.js/issues/19042


rwjblue:
Confirmed locally in the console that there are no deprecations when running the Ember tests until 3.20.

PR #297, opened 2020-08-10
Add version range to ember-cli-typescript deps

### LINKED ISSUE #296: Un-pin version of ember-cli-typescript
Hi, is it required to have the `ember-cli-typescript` pinned to a specific version? It would be better to use a version range to allow for a deps flattening.

https://github.com/glimmerjs/glimmer.js/blob/4f1029e7a87089aeff36627ad821c1ce1ce2b27e/packages/%40glimmer/component/package.json#L40

ondrejsevcik:
This fixes #296


rwjblue:
Will try to release shortly...

PR #298, opened 2020-08-10
feat: Throw better errors when upvars are missing

chadhietala:
This adds more actionanle error messages to the `CompileTimeResolver`
for modifiers and helpers. Prior to this, you would end up getting
cryptic errors around using undefined to as key into a Weakmap.


rwjblue:
Changes look good to me! Haven't dug into why CI is failing though the failures do seem related...

PR #301, opened 2020-09-02
[FEAT] Upgrade the VM to v0.61.1

pzuraq:
Upgrades the VM to the latest version

### DIFF
--- a/packages/@glimmer/core/test/non-interactive/render-test.ts
+++ b/packages/@glimmer/core/test/non-interactive/render-test.ts
@@ -281,7 +281,7 @@ module(`[@glimmer/core] non-interactive rendering tests`, () => {
       } catch (err) {
         assert.ok(
           err.message.match(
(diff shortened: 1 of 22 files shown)


packages/@glimmer/core/test/non-interactive/render-test.ts
```diff
@@ -281,7 +281,7 @@ module(`[@glimmer/core] non-interactive rendering tests`, () => {
       } catch (err) {
         assert.ok(
           err.message.match(
-            "You tried to reference {{name}} from the template-only-component template, which doesn't have an associated component class. Template-only components can only access args passed to them. Did you mean {{@name}}?"
+            'You attempted to access `this` on a template only component, template-only-component. Template only components do not have a `this` context, and can only access arguments'
```
rwjblue, line 284: Hmm, why do we loose the nicer error? If this is due to an upstream change, we should work to fix it...

pzuraq: it is, and it's noted in the location of the error with a link to this issue: https://github.com/glimmerjs/glimmer-vm/issues/1153

PR #315, opened 2020-10-26
Add support for partial rehydration and example apps for rehydration

chiragpat:
**Note**: In draft state until a new version of glimmer-vm is published

**Why**
The new version of glimmer-vm support partials rehydration. (see glimmerjs/glimmer-vm#1179). This PR bumps glimmer-vm versions and adds example applications to see how this feature can be used

**How**
- Bump to latest glimmer-vm
- Modify webpack.config to support server bundles from example apps
- Add new example apps for rehydration and partial rehydration
- Add unit test for partial rehydration

### DIFF
--- a/packages/@glimmer/component/package.json
+++ b/packages/@glimmer/component/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@glimmer/env": "^0.1.7",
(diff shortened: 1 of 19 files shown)


chadhietala:
Released 0.62.4 of the VM

PR #317, opened 2020-11-13
[FEAT] Upgrade to Glimmer VM v0.65.0

pzuraq:


rwjblue:
@pzuraq - Couple of failing tests for the babel plugins. Mind taking a look?

packages/@glimmer/core/src/managers/component/template-only.ts
```diff
@@ -1,10 +1,8 @@
 import {
```
rwjblue, line 1: Just curious, this whole file / concept should be upstreamed right?

pzuraq: Yes, it will be after a few more refactors

rwjblue (approved):

PR #318, opened 2020-11-16
Intersection observer

chadhietala:
This PR adds an example showing how to use intersection observer in
conjunction with partial rehydration.


packages/example-apps/partial-rehydration/index.ts
```diff
@@ -1,14 +1,36 @@
...
+rehydrate({
+  get RehydratableCounter() {
+    // Can load components async
+    return Promise.resolve(RehydratableCounter);
   },
-  { once: true }
-);
+});
+
+function rehydrate(componentMapping) {
```
rwjblue, line 11:
```suggestion
function rehydrate(componentMapping): void {
```

packages/example-apps/partial-rehydration/src/RehydratableRegion.ts
```diff
@@ -0,0 +1,17 @@
+import { createTemplate, setComponentTemplate, templateOnlyComponent } from '@glimmer/core';
+
+function toJSON(args) {
```
rwjblue, line 3:
```suggestion
function toJSON(args): string {
```

rwjblue (approved):

PR #324, opened 2020-12-21
Updating glimmer-vm to 0.73.0

chiragpat:
**How**
- Removed implementations and tests of things that have been upstreamed into the vm
- The resolver returns null for all methods as we are using strict mode which will handle all resolution logic
- Updated babel-plugin to pass in locals into the precompile step


packages/@glimmer/core/src/render-component/resolvers.ts
```diff
@@ -1,132 +1,63 @@
...
 /**
- * The RuntimeResolver is what is used to resolve everything. It is responsible
- * for registering root components (passed to `renderComponent`), and resolving
- * all other types of resolvables.
- *
- * The CompileTimeResolver is responsible for registering everything but root
- * components, which is why `registry` is public, for ease of access.
+ * Resolution for non built ins is now handled by the vm as we are using strict mode
  */
 export class RuntimeResolver implements VMRuntimeResolver {
```
pzuraq, line 15: We should look into making these optional now tbh

pzuraq: https://github.com/glimmerjs/glimmer-vm/issues/1250

packages/@glimmer/core/src/template.ts
```diff
@@ -1,92 +1,25 @@
...
 // This is provided by the `babel-plugin-strict-template-precompile` plugin
 export let createTemplate: (
   scopeOrTemplate: Dict<unknown> | string,
   template?: string
-) => CustomSerializedTemplate;
+) => SerializedTemplateWithLazyBlock;
 
 if (DEBUG) {
-  createTemplate = (): CustomSerializedTemplate => {
+  createTemplate = (): SerializedTemplateWithLazyBlock => {
```
pzuraq, line 13: We should probably rename/alias this to `precompileTemplate`, to match the original RFC. We can do that in a follow on, just noting for the 2.0 release.

pzuraq: https://github.com/glimmerjs/glimmer.js/issues/325

packages/@glimmer/core/src/template.ts
```diff
@@ -1,92 +1,25 @@
...
-
-    pointer = getPrototypeOf(pointer);
-  }
-
-  return undefined;
+export function setComponentTemplate(
+  template: SerializedTemplateWithLazyBlock,
+  ComponentClass: object
+): object {
+  return vmSetComponentTemplate(templateFactory(template), ComponentClass);
```
pzuraq, line 24: we should likely update babel transforms to make this unnecessary, the VM version matches the public API version in Ember, so it'd be good to make sure they stay aligned. Again, fine to do in a followup, and if the motivation is to save bytes we could expose a different compile-target function that does this.

pzuraq: https://github.com/glimmerjs/glimmer.js/issues/326

pzuraq (approved):

PR #328, opened 2021-01-06
Rename createTemplate to precompileTemplate to match spec defined in rfc

### LINKED ISSUE #325: Rename `createTemplate` to `precompileTemplate`
And match up the API with the one spec'd in the strict mode RFC, including implementation detail changes: https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md

chiragpat:
Resolves #325


locks:
The source code has since been renamed!

Issue #316, opened 2020-11-04
1.x emits invalid ES modules on Ember >= 3.20-beta.4


ef4:
`embroider-compat-audit` flagged the behavior implemented in https://github.com/glimmerjs/glimmer.js/pull/293 as a problem under ember >= 3.20.0-beta.4, because `ember-component-manager.ts` imports `setDestroyed` and `setDestroying` from `destroyables.ts`, but the overridden version of `destroyables.ts` doesn't export those names.

(This is consistent with the ES module spec, which considers importing a non-exported name a SyntaxError.)

It seems like it would be better to always present a consistent interface out of `destroyables.ts`, instead of having `ember-component-manager.ts` do its own separate `gte('3.20.0-beta.4')`. 

Alternatively, to keep things organized the same, we could switch to:

```js
import * as destroyables from './destroyables';
const { setDestroyed, setDestroying } = destroyables;
```

which is legal because it always imports the whole module namespace and then inspect it at runtime, where it may find that these names are undefined and that's OK. For searchability, the audit failure looks like:

<img width="687" alt="Screen Shot 2020-11-04 at 12 51 40 PM" src="https://user-images.githubusercontent.com/319282/98150357-8466e780-1e9c-11eb-8526-625697c1cfc1.png">

```
./node_modules/@glimmer/component/-private/ember-component-manager.js
  importing a non-existent named export: "./destroyables" has no export named "setDestroyed".
       6 | import { schedule } from '@ember/runloop';
       7 | import BaseComponentManager from './base-component-manager';
    >  8 | import { setDestroyed, setDestroying } from './destroyables';
         |         ^^^^^^^^^^^^
       9 | const CAPABILITIES = true // @ts-ignore
      10 | // @ts-ignore
      11 | ? capabilities('3.13', {
  importing a non-existent named export: "./destroyables" has no export named "setDestroying".
       6 | import { schedule } from '@ember/runloop';
       7 | import BaseComponentManager from './base-component-manager';
    >  8 | import { setDestroyed, setDestroying } from './destroyables';
         |                       ^^^^^^^^^^^^^
       9 | const CAPABILITIES = true // @ts-ignore
      10 | // @ts-ignore
      11 | ? capabilities('3.13', {
  file was included because:
    ./node_modules/@glimmer/component/index.js
    ./components/addons/ember-animated.js
    ./tests/integration/components/addons/ember-animated-test.js
    ./assets/test.js
    ./tests/index.html
[...]

PR #331, opened 2021-02-25
fix invalid imports

ef4:
This fixes https://github.com/glimmerjs/glimmer.js/issues/316.

That issue turns out to be more than an annoyance for Embroider's auditing code. As I'm experimenting with alternative packagers, I'm finding others that blow up because of the invalid use of imports here.

### DIFF
--- a/packages/@glimmer/component/addon/-private/ember-component-manager.ts
+++ b/packages/@glimmer/component/addon/-private/ember-component-manager.ts
@@ -11,7 +11,8 @@ import BaseComponentManager, {
 } from './base-component-manager';
 
 import GlimmerComponent, { Constructor } from './component';
-import { setDestroyed, setDestroying } from './destroyables';
+import * as destroyables from './destroyables';
+const { setDestroyed, setDestroying } = destroyables;
 
 const CAPABILITIES = gte('3.13.0-beta.1')
   ? capabilities('3.13', {


ef4:
It would be great if we could release this as 1.0.4.

rwjblue (approved):

packages/@glimmer/component/addon/-private/ember-component-manager.ts
```diff
@@ -11,7 +11,8 @@ import BaseComponentManager, {
 } from './base-component-manager';
 
 import GlimmerComponent, { Constructor } from './component';
-import { setDestroyed, setDestroying } from './destroyables';
+import * as destroyables from './destroyables';
+const { setDestroyed, setDestroying } = destroyables;
```
Turbo87, line 15: I'm not sure I understand this change. shouldn't these two be equivalent? 🤔

ef4: The original is a syntax error if `./destroyables` doesn't actually export either of those names. That happens in our implementation due to https://github.com/glimmerjs/glimmer.js/blob/8db4a62d1714bb47d15c09f3809cd8111cc07d43/packages/%40glimmer/component/ember-addon-main.js#L59-L63.

The new code avoids that problem by importing the whole namespace, and destructuring it *at runtime*. 

You can see the difference if you try an example directly in a browser.

Turbo87: I see, thanks 👍

PR #333, opened 2021-03-16
Update Glimmer VM to 0.77.1

pzuraq:


packages/@glimmer/component/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@glimmer/env": "^0.1.7",
-    "@glimmer/util": "0.73.0",
+    "@glimmer/util": "0.77.1",
```
rwjblue, line 30: I think 0.77.2 is published, and we should probably use the latest unless there is some reason we can't...

PR #343, opened 2021-04-12
Add support for components using private methods

chadhietala:
@babel/plugin-proposal-private-methods requires seeing
@babel/plugin-proposal-class-properties being set before it can run. In
Babel plugins run before presets. Because of this, these 2 plugins must
exist in the preset.

### DIFF
--- a/packages/@glimmer/babel-preset/index.js
+++ b/packages/@glimmer/babel-preset/index.js
@@ -11,6 +11,13 @@ module.exports = function (api, options) {
 
   let precompile, templateCompilerPath;
 
+  const looseProps =
+    typeof options.loose === 'object' &&
+    options.loose !== null &&
+    typeof options.loose.properties === 'boolean'
+      ? options.loose.properties
+      : false;
+
   if (options.precompile) {
     precompile = options.precompile;
   } else {
@@ -75,9 +82,19 @@ module.exports = function (api, options) {
         { legacy: true },
       ],
 
-      __loadPlugins
-        ? require('@babel/plugin-proposal-class-properties')
-        : require.resolve('@babel/plugin-proposal-class-properties'),
+      [
+        __loadPlugins
+          ? require('@babel/plugin-proposal-class-properties')
+          : require.resolve('@babel/plugin-proposal-class-properties'),
+        { loose: looseProps },
+      ],
+
+      [
+        __loadPlugins
+          ? require('@babel/plugin-proposal-private-methods')
+          : require.resolve('@babel/plugin-proposal-private-methods'),
+        { loose: looseProps },
+      ],
     ],
   };
 };
--- a/packages/@glimmer/babel-preset/package.json
+++ b/packages/@glimmer/babel-preset/package.json
@@ -7,10 +7,11 @@
   "license": "MIT",
   "private": false,
   "scripts": {
-    "test": "mocha -r esm"
+    "test": "mocha -r esm --timeout 10000"
   },
   "dependencies": {
     "@babel/plugin-proposal-class-properties": "^7.8.3",
+    "@babel/plugin-proposal-private-methods": "^7.13.0",
     "@babel/plugin-proposal-decorators": "^7.8.3",
     "@glimmer/compiler": "0.77.6",
     "@glimmer/vm-babel-plugins": "0.77.6",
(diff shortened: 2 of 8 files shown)


packages/@glimmer/babel-preset/index.js
```diff
@@ -75,9 +75,19 @@ module.exports = function (api, options) {
...
       ],
 
-      __loadPlugins
-        ? require('@babel/plugin-proposal-class-properties')
-        : require.resolve('@babel/plugin-proposal-class-properties'),
+      [
+        __loadPlugins
+          ? require('@babel/plugin-proposal-class-properties')
+          : require.resolve('@babel/plugin-proposal-class-properties'),
+        { loose: true },
```
pzuraq, line 82: I think we should make this configurable, and default it to `false`. Relying on loose mode semantics is not a good idea for future proofing, and this should generally be discouraged.

pzuraq (changes requested):

packages/@glimmer/babel-preset/index.js
```diff
@@ -75,9 +75,19 @@ module.exports = function (api, options) {
...
+      ],
+
+      [
+        __loadPlugins
+          ? require('@babel/plugin-proposal-private-methods')
+          : require.resolve('@babel/plugin-proposal-private-methods'),
+        { loose: true },
```
pzuraq, line 89: Same as above

chadhietala: This is going to emit `WeakSet` in strict mode. Are we ok with that?

rwjblue: Not really, I don't think we can use `WeakSet` (IE11 doesn't support it and we don't directly polyfill the global).

chadhietala: Ok well then this is going to have to default to `loose: true` because these plugins need to [share the options](https://babeljs.io/docs/en/babel-plugin-proposal-private-methods#loose).

rwjblue:
Do we actually use private methods? Do we have to?

chadhietala:
I'm trying to migrate to this and was supporting private methods. The issue is that babel plugins run before presets and the private method plugin depends on the class properties plugin.

Now it would be "fine" to have these as both plugins but the issue is that when the preset runs the fields have already been transformed and thus presets can't find `static template =` to transforming it. For example this is the current output if you both set as plugins:

Input:

```js
  import { Hello } from "../Hello.js";
  import Component, { hbs } from '@glimmerx/component';
  export default class MyPage extends Component {
    static template = hbs\`
      <Hello />
    \`
  }
```

Current Output:

```js
class MyPage extends _glimmerx_component__WEBPACK_IMPORTED_MODULE_2__.default {}
MyPage.template = (0,_glimmer_core__WEBPACK_IMPORTED_MODULE_0__.setComponentTemplate)((0,_glimmer_core__WEBPACK_IMPORTED_MODULE_0__.createTemplateFactory)(
/*
          <Hello />
*/
{
  "id": "Ng5WBitS",
  "block": "[[[1,\"\\n          \"],[8,[32,0],null,null,null],[1,\"\\n        \"]],[],false,[]]",
  "moduleName": "(unknown template module)",
  "scope": () => [_var_folders_0_6vsqqq293r1_rddztq1585xc0008c7_T_broccoli_97650kjsjBvbcT6M6_out_2_broccoli_persistent_filter_babel__WEBPACK_IMPORTED_MODULE_1__.Hello],
  "isStrictMode": true
}), (0,_glimmer_core__WEBPACK_IMPORTED_MODULE_0__.templateOnlyComponent)("MyPage", "_MyPage"));
```

You could do a thing where you have plugins that re-transform transformed code but that seems like a slippery slope.

packages/@glimmer/babel-preset/index.js
pzuraq: We can still default it to `false` and then, if you need to support IE11, allow users to set it to `true`. I'm, very hesitant to say the default is something that _will_ cause breaking failures when we eventually are no longer polyfilling (we have actually had failures caused by this both in Ember and elsewhere).

chadhietala: I'm not really sure how to move forward here.

rwjblue (commented):
Chatted with @chadhietala and @pzuraq, and I think the path forward here is to remove `{ loose: true }` (and use strict mode) and then we can land/release.

rwjblue (approved):

PR #358, opened 2021-08-24
Adds @cached decorator

NullVoxPopuli:
This is the same implementation as ember and the polyfill

### DIFF
--- a/packages/@glimmer/tracking/src/cached.ts
+++ b/packages/@glimmer/tracking/src/cached.ts
@@ -1,6 +1,63 @@
 import { DEBUG } from '@glimmer/env';
 import { createCache, getValue } from '@glimmer/validator';
 
+/**
+ * @decorator
+ *
+ * The `@cached` decorator can be used on getters in order to cache the return
+ * value of the getter. This is useful when a getter is expensive and used very
+ * often.
+ *
+ *
+ * @example
+ *
+ * in this guest list class, we have the `sortedGuests`
+ * getter that sorts the guests alphabetically:
+ *
+ * ```js
+ * import { tracked } from '@glimmer/tracking';
+ *
+ * class GuestList {
+ *   @tracked guests = ['Zoey', 'Tomster'];
+ *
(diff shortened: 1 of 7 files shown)


snewcomer (approved):
https://github.com/emberjs/rfcs/pull/566

Beautiful!

packages/@glimmer/tracking/src/cached.ts
```diff
@@ -0,0 +1,108 @@
...
+ * getter that sorts the guests alphabetically:
+ *
+ * ```js
+ * import { tracked } from '@glimmer/tracking';
+ * class GuestList {
+ *   @tracked guests = ['Zoey', 'Tomster'];
+ *   get sortedGuests() {
+ *     return this.guests.slice().sort()
+ *   }
+ * }
```
knownasilya, line 24: I'd recommend being generous with empty lines, since this will be read a bit. So a new-line after the import, and one after the guests definition.

pzuraq:
I think this needs to be implemented in Ember as well, since Ember has its own implementation of `@glimmer/tracking` for the moment still

NullVoxPopuli:
Ember version: https://github.com/emberjs/ember.js/pull/19772

mydea:
Since this has landed in ember, would it be possible to merge this as well? As of now, e.g. when using Typescript it is a bit akward, since `cached` types do not exist in `@glimmer/tracking` but can actually be imported from that path :sweat_smile:

rwjblue:
Ya, we should definitely get this landed. 

@NullVoxPopuli - Can you rebase to kick off a new set of tests?

NullVoxPopuli:
it's already up to date / fully rebased

NullVoxPopuli:
@rwjblue any way you/we/i can click the "re-run all jobs button"?

rwjblue:
Hmm, no? There is no "re-run all" button on the checks tab.

knownasilya:
🎉

PR #360, opened 2021-09-10
Fix type errors on ember canary

NullVoxPopuli:
This should hopefully improve the ember-beta and ember-canary CI

PR #361, opened 2021-09-13
Drop support for Node 10

NullVoxPopuli:
While trying to fix some things in this PR: https://github.com/glimmerjs/glimmer.js/pull/360/checks?check_run_id=3571369240
I noticed that ember no longer supports Node 10

idk if this is the right thing to do -- just trying to unblock PRs atm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10, 12, 13]
+        node-version: [12, 13]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10, 12, 13]
+        node-version: [12, 13]
```
rwjblue, line 28: Node 13 is itself unsupported, so we should drop that too. I'd like to see us go to this:

```suggestion
        node-version: [12, 14, 16]
```

PR #371, opened 2021-12-21
Updated App.js for accessibility.

MelSumner:
Updated the output for App.js so it won't fail accessibility checkers out of the box.


dgeb (approved):
Thanks @MelSumner 👍

PR #385, opened 2022-03-23
Component signature

chriskrycho:
Adds backwards-compatible support for component `Signature`s per [Ember RFC #0748][rfc]. The public API of `Component` is now *more* permissive than it was previously, because it changes the type parameter to be an unconstrained generic `<S>` (for "signature") which can accept *either* the existing `Args` types *or* a new `Signature` which includes `Args` and adds `Blocks` and `Element`.

[rfc]: https://github.com/emberjs/rfcs/pull/748

The `Args` part of the new signature work exactly like the old args-only type did. The `Blocks` and `Element` parts of a signature are inert from the perspective of TypeScript users who are not yet using [Glint][glint], but Glint users will benefit directly once Glint releases an update which can requires a version of `@glimmer/component` incorporating this change.

[glint]: https://github.com/typed-ember/glint

Since this change is backwards compatible, we can deprecate the non-`Signature` form at a later time when we are ready for a 2.0 release.

To validate these changes, with the relatively complicated type machinery they require under the hood, this also introduces the `expect-type` type testing library, rewrites the existing type tests using it, and introduces new type tests for all supported forms of the `Signature`.

### DIFF
--- a/packages/@glimmer/component/addon/-private/component.ts
+++ b/packages/@glimmer/component/addon/-private/component.ts
@@ -10,11 +10,65 @@ export function setDestroyed(component: GlimmerComponent<object>): void {
   DESTROYED.set(component, true);
 }
 
-export let ARGS_SET: WeakMap<object, boolean>;
+// This provides a type-safe `WeakMap`: the getter and setter link the key to a
+// specific value. This is how `WeakMap`s actually behave, but the TS type
+// system does not (yet!) have a good way to capture that for types like
+// `WeakMap` where the type is generic over another generic type (here, `Args`).
+interface ArgsSetMap extends WeakMap<Args<unknown>, boolean> {
+  get<S>(key: Args<S>): boolean | undefined;
+  set<S>(key: Args<S>, value: boolean): this;
+}
+
+// SAFETY: this only holds because we *only* acces this when `DEBUG` is `true`.
+// There is not a great way to connect that data in TS at present.
+export let ARGS_SET: ArgsSetMap;
 
 if (DEBUG) {
-  ARGS_SET = new WeakMap();
+  ARGS_SET = new WeakMap() as ArgsSetMap;
 }
+
+// --- Type utilities for component signatures --- //
+
+// This provides us a way to have a "fallback" which represents an empty object,
+// without the downsides of how TS treats `{}`. Specifically: this will
+// correctly leverage "excess property checking" so that, given a component
+// which has no named args, if someone invokes it with any named args, they will
+// get a type error.
+declare const Empty: unique symbol;
+type EmptyObject = { [Empty]?: true };
+
+type GetOrElse<Obj, K, Fallback> = K extends keyof Obj ? Obj[K] : Fallback;
+
+/** Given a signature `S`, get back the `Args` type. */
+type ArgsFor<S> = 'Args' extends keyof S
+  ? S['Args'] extends { Named?: object; Positional?: unknown[] } // Are they longhand already?
+    ? {
+        Named: GetOrElse<S['Args'], 'Named', EmptyObject>;
+        Positional: GetOrElse<S['Args'], 'Positional', []>;
+      }
+    : { Named: S['Args']; Positional: [] }
+  : { Named: EmptyObject; Positional: [] };
+
+/** Given any allowed shorthand form of a signature, desugars it to its full expanded type */
+type ExpandSignature<T> = {
+  Element: GetOrElse<T, 'Element', null>;
+  Args: keyof T extends 'Args' | 'Element' | 'Blocks' // Is this a `Signature`?
+    ? ArgsFor<T> // Then use `Signature` args
+    : { Named: T; Positional: [] }; // Otherwise fall back to classic `Args`.
+  Blocks: 'Blocks' extends keyof T
+    ? {
+        [Block in keyof T['Blocks']]: T['Blocks'][Block] extends unknown[]
+          ? { Positional: T['Blocks'][Block] }
+          : T['Blocks'][Block];
+      }
+    : EmptyObject;
+};
+
+/**
+ * @internal we use this type for convenience internally; inference means users
+ *   should not normally need to name it
+ */
+export type Args<S> = ExpandSignature<S>['Args']['Named'];
 
 /**
  * The `Component` class defines an encapsulated UI element that is rendered to
@@ -139,7 +193,7 @@ if (DEBUG) {
  * `args` property. For example, if `{{@firstName}}` is `Tom` in the template,
  * inside the component `this.args.firstName` would also be `Tom`.
  */
-export default class GlimmerComponent<Args extends {} = {}> {
+export default class GlimmerComponent<S = unknown> {
   /**
    * Constructs a new component and assigns itself the passed properties. You
    * should not construct new components yourself. Instead, Glimmer will
@@ -148,7 +202,7 @@ export default class GlimmerComponent<Args extends {} = {}> {
    * @param owner
    * @param args
    */
-  constructor(_owner: unknown, args: Args) {
+  constructor(_owner: unknown, args: Args<S>) {
     if (DEBUG && !ARGS_SET.has(args)) {
       throw new Error(
         `You must pass both the owner and args to super() in your component: ${this.constructor.name}. You can pass them directly, or use ...arguments to pass all arguments through.`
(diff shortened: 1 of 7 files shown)


chadhietala:
LGTM

packages/@glimmer/component/addon/index.ts
```diff
@@ -12,11 +12,11 @@ declare module '@ember/component' {
 import { setComponentManager } from '@ember/component';
 
 import GlimmerComponentManager from './-private/ember-component-manager';
-import _GlimmerComponent from './-private/component';
+import _GlimmerComponent, { Args } from './-private/component';
 import { setOwner } from '@ember/application';
 
-export default class GlimmerComponent extends _GlimmerComponent {
-  constructor(owner, args) {
+export default class GlimmerComponent<S> extends _GlimmerComponent<S> {
```
jamescdavis, line 18: Unless we provide a default for `S` here, this _requires_ users to specify a type arg, which is a breaking change. In fact, many tests no longer type-check with this change (try running `tsc` in the repo root).

packages/@glimmer/component/addon/index.ts
```diff
@@ -12,11 +12,11 @@ declare module '@ember/component' {
...
+  constructor(owner: object, args: Args<S>) {
```
jamescdavis, line 19: Why is `owner` `object` here and not `unknown` as it is in the parent?

packages/@glimmer/component/addon/index.ts
chriskrycho: Ahhh, good catch. I thought I’d caught this everywhere I did it, but: no. Thank you!

packages/@glimmer/component/addon/index.ts
chriskrycho: A good question, and I had the same one. It's because `setOwner` takes an `object` argument. It’s very likely something we want to rationalize throughout… but not in *this* PR.

packages/@glimmer/component/addon/index.ts
chriskrycho:
```suggestion
export default class GlimmerComponent<S = unknown> extends _GlimmerComponent<S> {
```

chriskrycho: I fixed it here, but/and I'm concerned that CI was passing without `tsc` checking! Will follow up on that front tomorrow.

test/types/component-test.ts
```diff
@@ -1,39 +1,96 @@
...
+expectTypeOf(componentWithArgsOnly.args).toEqualTypeOf<Readonly<Args>>();
+
+interface ElementOnly {
+  Element: HTMLParagraphElement;
+}
+
+const componentWithElOnly = new Component<ElementOnly>({}, {});
+
+// We cannot check on toEqualTypeOf here b/c EmptyObject is intentionally not
+// public.
```
dfreeman, line 37: Could this cause issues with declaration emit? I'm not sure how likely this actually is to come up in practice, but as a super contrived example, if I were to write something like:

```ts
class MyComponent extends Component<{}> {
  foo = this.args;
}
```

I think `tsc` would fail trying to produce a type annotation for `foo` for a `.d.ts` file, wouldn't it?

packages/@glimmer/component/addon/-private/component.ts
```diff
@@ -10,12 +10,66 @@ export function setDestroyed(component: GlimmerComponent<object>): void {
...
+  ? S['Args'] extends { Named?: object; Positional?: unknown[] } // Are they longhand already?
+    ? {
+        Named: GetOrElse<S['Args'], 'Named', EmptyObject>;
+        Positional: GetOrElse<S['Args'], 'Positional', []>;
+      }
+    : { Named: S['Args']; Positional: [] }
+  : { Named: EmptyObject; Positional: [] };
+
+/** Given any allowed shorthand form of a signature, desugars it to its full expanded type */
+type ExpandSignature<T> = {
```
dfreeman, line 53: I wonder if there's value (or rather, if the value outweighs the hazard) in exporting some or all of these types as `@internal`.

Otherwise we'll have to duplicate them in Glint, which certainly isn't the end of the world, but there's a certain irony to the fact that this package doesn't even _do_ anything with `Element` or `Blocks` and yet contains the canonical expansion logic for them 😄

dfreeman (commented):
This looks great; thanks for getting it implemented @chriskrycho! Just a couple questions around type visibility that the answer to might be 'nope, this is exactly how we want it already'

packages/@glimmer/component/addon/-private/component.ts
chriskrycho: Exporting as `@internal` makes a lot of sense to me. I had the same thought when authoring it.

jamescdavis: +1

packages/@glimmer/component/src/component.ts
```diff
@@ -1,10 +1,10 @@
 import { setComponentManager, setOwner } from '@glimmer/core';
 import GlimmerComponentManager from './component-manager';
-import _GlimmerComponent from '../addon/-private/component';
+import _GlimmerComponent, { Args } from '../addon/-private/component';
 import { DEBUG } from '@glimmer/env';
 
-export default class GlimmerComponent<Args extends {} = {}> extends _GlimmerComponent<Args> {
-  constructor(owner: object, args: Args) {
+export default class GlimmerComponent<S> extends _GlimmerComponent<S> {
```
jamescdavis, line 6: Oops, it's actually this one that needs a default because:

https://github.com/glimmerjs/glimmer.js/blob/ee9fbf09fdf012a512f0ff47872afe3a05a5d51b/packages/%40glimmer/component/index.ts#L1

```suggestion
export default class GlimmerComponent<S = unknown> extends _GlimmerComponent<S> {
```

packages/@glimmer/component/addon/index.ts
jamescdavis: We've [taught people](https://docs.ember-cli-typescript.com/ember/components#adding-arguments) to do:
```ts
export default class MyComponent extends Component<Args> {
  constructor(owner: unknown, args: Args) {
    super(owner, args);
    ⋮
  }
}
```
So making this `object` here will force users update their constructor type signatures (otherwise the `super()` call doesn't type check: `Argument of type 'unknown' is not assignable to parameter of type 'object'.`).

test/types/component-test.ts
chriskrycho: Good call. I’ll export this with an `@internal` annotation… and we’ll need to treat its *location* as public API for imports so existing declarations don’t break.

chriskrycho: Resolved via 6fc4a17!

packages/@glimmer/component/src/component.ts
chriskrycho: Ahhh, thank you.

packages/@glimmer/component/addon/index.ts
chriskrycho: The bit with `unknown` is actually broken today, so we need to update our docs—we have (somewhere!) been publishing with requiring `object` *anyway*, so folks have been needing to introduce dummy `Owner` object blobs.

chriskrycho:
Note for myself: looks like once we land this I'll need to back-port it to the v1.x branch so we can cut a v1.1 release.

packages/@glimmer/component/addon/-private/component.ts
```diff
@@ -28,14 +28,25 @@ if (DEBUG) {
...
+ * correctly leverage "excess property checking" so that, given a component
+ * which has no named args, if someone invokes it with any named args, they will
+ * get a type error.
+ *
+ * @internal This is exported so declaration emit works (if it were not emitted,
+ *   declarations which fall back to it would not wor). It is *not* intended for
+ *   public usage, and the specific mechanics it uses may change at any time.
+ *   The location of this export *is* part of the public API, because moving it
+ *   will cause existing declarations, but is not legal for end users to import
+ *   themselves, so ***DO NOT RELY ON IT***.
```
dfreeman, line 47:
```suggestion
 * @internal This is exported so declaration emit works (if it were not emitted,
 *   declarations which fall back to it would not work). It is *not* intended for
 *   public usage, and the specific mechanics it uses may change at any time.
 *   The location of this export *is* part of the public API, because moving it
 *   will break existing declarations, but is not legal for end users to import
 *   themselves, so ***DO NOT RELY ON IT***.
```

PR #391, opened 2022-04-01
Fix Signature handling for unions and generics

chriskrycho:
Cherry-picks this fix to `master` from #390 (which targets `v1.x`).

Ultimately, we should be able to drop these shenanigans entirely before cutting v2.0, but we’ll want to give folks currently using the v2 betas a heads-up about it.

### DIFF
--- a/test/types/component-test.ts
+++ b/test/types/component-test.ts
@@ -1,5 +1,9 @@
 import { expectTypeOf } from 'expect-type';
+
+// Intentionally checking the shape of the exports *and* the export itself.
 import * as gc from '@glimmer/component';
+// tslint:disable-next-line: no-duplicate-imports
+import Component from '@glimmer/component';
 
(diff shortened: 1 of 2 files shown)


test/types/component-test.ts
```diff
@@ -1,37 +1,66 @@
 import { expectTypeOf } from 'expect-type';
+
+// Intentionally checking the shape of the exports *and* the export itself.
 import * as gc from '@glimmer/component';
+// eslint-disable-next-line no-duplicate-imports
+import Component from '@glimmer/component';
```
chadhietala, line 6: Linting doesn't seem to like this

chriskrycho: Ahhh, the problem is that it's tslint instead of eslint. Ugh.

chadhietala (approved):

PR #392, opened 2022-04-08
Fix `Signature` integration for external consumers

dfreeman:
There are two hazards currently facing consumers of the mostly-private `ExpandSignature` type that landed in #385.
 - The [Signature RFC](https://github.com/emberjs/rfcs/pull/748) called for a `Params` wrapper around the fully-expanded form of a block definition, which is missing here when the user provides the shorthand
 - v1 and v2 of `@glimmer/component` publish their types in different locations (alone in `dist/types/...` in v1, but alongside the implementation in`dist/{modules,commonjs}/...` in v2)

This PR adds the missing `{ Params: ... }` wrapper and a `typesVersions` entry that will allow for a standardized import path between v1 and v2 for privileged consumers of the `ExpandSignature` type.

I'll also plan to open an equivalent PR to this one against the `v1.x` branch, assuming everything looks ok here.

/cc @chriskrycho

### DIFF
--- a/packages/@glimmer/component/addon/-private/component.ts
+++ b/packages/@glimmer/component/addon/-private/component.ts
@@ -68,7 +68,7 @@ type _ExpandSignature<T> = {
   Blocks: 'Blocks' extends keyof T
     ? {
         [Block in keyof T['Blocks']]: T['Blocks'][Block] extends unknown[]
-          ? { Positional: T['Blocks'][Block] }
+          ? { Params: { Positional: T['Blocks'][Block] } }
           : T['Blocks'][Block];
       }
     : EmptyObject;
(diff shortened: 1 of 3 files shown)


packages/@glimmer/component/addon/-private/component.ts
```diff
@@ -68,7 +68,7 @@ type _ExpandSignature<T> = {
   Blocks: 'Blocks' extends keyof T
     ? {
         [Block in keyof T['Blocks']]: T['Blocks'][Block] extends unknown[]
-          ? { Positional: T['Blocks'][Block] }
+          ? { Params: { Positional: T['Blocks'][Block] } }
```
chriskrycho, line 71: Thanks for fixing this (and my bad for missing it)!

chriskrycho (approved):
💙 thanks!

PR #393, opened 2022-04-08
Backport: Fix `Signature` integration for external consumers

dfreeman:
This backports #392 to `v1.x`.

The only notable difference is that because of the way types tests on the 1.x branch use `paths` to locate the published packages in `dist`, the type mapping has to exist both in the tests' `tsconfig.json` _and_ in `@glint/component`'s `package.json`.


chriskrycho (approved):
Again, thank you!

PR #412, opened 2024-03-19
Widen range on `@glimmer/validator` so that package-managers respecti…

NullVoxPopuli:
…ng semver can use a more up to date version of `@glimmer/validator`


NullVoxPopuli:
oh no, this repo uses yarn :cry:

